PR #596, opened 2024-07-16
fix(Switch): Only render label component if children is defined

TimAufdemKampe:


mfal (approved):
Looks good. Thanks for the improvement!

PR #599, opened 2024-07-18
fix(Text): Fix Text style if text is used without a wrapper component

TimAufdemKampe:


packages/components/src/components/Text/Text.tsx
```diff
@@ -32,7 +32,11 @@ export const Text = flowComponent("Text", (props) => {
     ...rest
   } = props;
 
-  const rootClassName = clsx(color && styles[color], className);
+  const rootClassName = clsx(
```
mfal, line 35: I don't think setting the color to "dark" might not work in every use case, since text should inherit the color of the parent element per default. Maybe you should always add a `.flow--text` class (to get the magic global Flow style) and add additional color classes. This CSS can be used:

```scss
.text {
  color: inherit;

  &:where(.light) {
    color: var(--content--color--light);
  }

  &:where(.dark) {
    color: var(--content--color--dark);
  }
}
```

mfal (changes requested):

mfal (approved):

PR #638, opened 2024-08-15
fix(Section): add popover trigger to section header props context

Lisa18289:


mfal (commented):
I think the ModalTrigger might have the same issue. Can you check it and  extend the PropsContext if so?

mfal (approved):

PR #727, opened 2024-09-16
feat(CodeBlock): add code block component

Lisa18289:


packages/components/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,47 @@
...
+import React from "react";
+import type { SyntaxHighlighterProps } from "react-syntax-highlighter";
+import SyntaxHighlighter from "react-syntax-highlighter";
+import type { PropsWithClassName } from "@/lib/types/props";
+import clsx from "clsx";
+import { CopyButton } from "@/components/CopyButton";
+import styles from "./CodeBlock.module.scss";
+import { a11yDark } from "react-syntax-highlighter/dist/esm/styles/prism";
+
+export interface CodeBlockProps
```
mfal, line 11: I would prefer a `code` property instead of using it as child, because i guess that in practice the code itself, will rarely be provided inline.

elmartinezc (approved):

Issue #784, opened 2024-09-25
build:icons script of components package fails


TimAufdemKampe:
The `build:icons` script fails with an error.
![Bildschirmfoto 2024-09-25 um 14 42 51](https://github.com/user-attachments/assets/546d23ea-3c4e-4e6c-bc07-34ccf001ceae)

maaaathis:
Which nodejs version are you using @TimAufdemKampe ?

It works for me with nodejs 20.17.0.

maaaathis:
> > Which nodejs version are you using @TimAufdemKampe ?
> > It works for me with nodejs 20.17.0. We need methods here that are introduced with nodejs 20.11.0: [nodejs.org/en/blog/release/v20.11.0](https://nodejs.org/en/blog/release/v20.11.0)
> 
> Oh okay, i'm using Node version 20.9.0. Version 20.17.0 works perfectly fine 👍
> 
> Should be require a node version by the `engines` property in the `package.json`? To prevent other contributors from getting the same problem. @mfal

Is already set to >=20.11: https://github.com/mittwald/flow/blob/main/package.json#L9

PR #785, opened 2024-09-25
fix(build:icons script): Fix error of build:icons script

TimAufdemKampe:


maaaathis:
Which nodejs version are you using? The old handling worked well for me: https://github.com/mittwald/flow/issues/784#issuecomment-2374005257

PR #942, opened 2024-11-05
fix(OverlayController): fix overlayController handler management

JLampeMW:
- Simplified handler management in OverlayController
- Removed separate onOpen/onClose properties in favor of unified handler sets
- No functional changes, just internal restructuring

### DIFF
--- a/packages/components/src/lib/controller/overlay/OverlayController.ts
+++ b/packages/components/src/lib/controller/overlay/OverlayController.ts
@@ -10,8 +10,8 @@ export interface OverlayControllerOptions {
 
 export class OverlayController {
   public isOpen: boolean;
-  public onOpen?: () => void;
-  public onClose?: () => void;
+  private onOpenHandlers = new Set<() => void>();
+  private onCloseHandlers = new Set<() => void>();
 
   public constructor(options?: OverlayControllerOptions) {
     makeObservable(this, {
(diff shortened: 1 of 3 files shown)


packages/components/src/lib/controller/overlay/useOverlayController.ts
```diff
@@ -25,7 +26,26 @@ export const useOverlayController = (
...
+    }
+
+    if (onClose && controller.isOpen) {
+      disposers.push(controller.addOnClose(onClose));
+    }
+
+    return () => {
+      disposers.forEach((dispose) => dispose());
+    };
+  }, [onOpen, onClose, controller.isOpen]);
```
mfal, line 48: The actual open state should have any effect on binding handlers.

```suggestion
  useEffect(() => {
    const disposers: (() => void)[] = [];

    if (onOpen) {
      disposers.push(controller.addOnOpen(onOpen));
    }

    if (onClose) {
      disposers.push(controller.addOnClose(onClose));
    }

    return () => {
      disposers.forEach((dispose) => dispose());
    };
  }, [onOpen, onClose]);
```

packages/components/src/lib/controller/overlay/OverlayController.ts
```diff
@@ -10,8 +10,8 @@ export interface OverlayControllerOptions {
 
 export class OverlayController {
```
mfal, line 11: Maybe you should introduce two new local types for the handler and disposer to clarify the intent of this functions.

```typescript
type OpenStateHandler = () => void;
type DisposerFn = () => void;
```

mfal (changes requested):

mfal (approved):

PR #976, opened 2024-11-15
fix/style: adjusted the primary buttons position

BenniEngel:
adjusted the primary buttons position if it is the only one in the action group


mfal (changes requested):
This does not work, if there are additional secondary buttons.

See Section "Zusätzlicher Secondary Button" in Style Guide.

Lisa18289 (approved):

mfal (approved):

PR #1088, opened 2024-12-18
feat: add remote component & element generation

maaaathis:


maaaathis:
Superset by https://github.com/mittwald/flow/pull/852/commits/178ca3cf63967046f1b293c728f45dd189998986

PR #1165, opened 2025-02-10
Fix/translation error

ins0:
fixes https://github.com/mittwald/flow/issues/1092


mfal:
@ins0 this PR is not needed anymore since there is another (I assume "default") solution that does not require any custom implementation. Thanks anyway 🙏

PR #1259, opened 2025-03-10
Add remote component tests

mfal:


packages/components/vitest.workspace.dev.ts
```diff
@@ -0,0 +1,11 @@
+import { defineWorkspace } from "vitest/config";
+
+export default defineWorkspace([
+  {
+    extends: "./vitest.config.ts",
+    test: {
+      include: ["dev/**/*.test.{ts,tsx}"],
+      environment: "node",
```
ins0, line 8: is it intendet that the env is different on dev and prod?

packages/remote-react-components/dev/vitest/remote-test-server/createTestServer.tsx
```diff
@@ -0,0 +1,36 @@
...
+          "Tetst server port is already in use:",
+          remoteTestServerPort,
+        );
+        testServer.stop();
+        return;
+      }
+      await server.listen(remoteTestServerPort);
+      server.printUrls();
+    },
+    stop: async () => {
```
ins0, line 30: it would be more consistent if the server is created beforehand in a variable and start would finally starts listen, so stop can be used as normal with the desired log entry.

PR #1299, opened 2025-03-20
feat(Button): support async submit without action

### LINKED ISSUE #1252: Button without action does not update state
When used outside of an Action, the Button does not update its state. The loading, success and failure states are not shown.

## Example
When using a Button with a TextField in an Align component with a Form that has an async submit action:
```tsx
<Section>
  <Form form={form} onSubmit={async () => await sleep(3000)}>
    <Align>
      <TextField />
      <Button type="submit">Submit</Button>
    </Align>
  </Form>
</Section>
```

ntdoJanneck:
closes #1252


ntdoJanneck:
new pull request: https://github.com/mittwald/flow/pull/1303

PR #1307, opened 2025-03-21
fix(Label): pick supported properties

Lisa18289:


packages/components/src/components/Label/Label.tsx
mfal, line 1: - Why is this required at all?
- The props are now limited, but in the component implementation all props are still picked as `...rest`.

Lisa18289: i thought it would be easier to just pick the properties we actually need. i could also omit the hidden property, or make it work, but their might be other properties that do not really work either.

mfal: Let's talk about this next time in office 😄

Issue #1437, opened 2025-05-08
Support `text-wrap` property for Text component


LukasFritzeDev:
It would be great to have the option to set the [CSS-Property `text-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap) on the `Text` and `Heading` component. Using this with the Value `balance` multiline Text looks much better.

[Blog post with detailed explanation](https://blog.stephaniestimac.com/posts/2023/10/css-text-wrap/)

### `Text` in `AccentBox`

without `text-wrap: balance`:
![Image](https://github.com/user-attachments/assets/4cf6db40-8c46-47c3-a1d2-ea48ee871f46)

with `text-wrap: balance`:
![Image](https://github.com/user-attachments/assets/375e59f9-44b6-4f85-b60d-9bfb8b78289f)

### `Text` in `IllustratedMessage`

without `text-wrap: balance`:
![Image](https://github.com/user-attachments/assets/6366074c-689a-4590-8a13-dae40dfc7975)

with `text-wrap: balance`:
![Image](https://github.com/user-attachments/assets/0426de3d-3669-47df-bf0c-6f989690ce5f)

---

Sample Implementation:

```tsx
    <ColumnLayout m={[2, 2]}>
      <AccentBox color="neutral">
        <Flex direction="column" gap="s" align="center" text-align="center">
          <BigNumber>
            <Text>{profile.issueSummary?.errors}</Text>
            <Text>Fehler</Text>
          </BigNumber>
            <Text align="center">
              kritische Fehler, die mit höchster Priorität umgesetzt werden
              sollten
            </Text>
        </Flex>
      </AccentBox>
      <AccentBox color="neutral">
        <Flex direction="column" gap="xs" align="center">
          <BigNumber>
            <Text>{profile.issueSummary?.warnings}</Text>
            <Text>Warnungen</Text>
          </BigNumber>
          <Text align="center">
            Punkte, die überprüft und bei Bedarf verbessert werden sollten
          </Text>
        </Flex>
      </AccentBox>
    </ColumnLayout>
```

mfal:
Although `balance` is a working draft, it is supported by the major browsers and does not have any negative impact, if not so. @LukasFritzeDev would you like to add this feature in a PR?

mfal:
Closed by #1446

PR #1446, opened 2025-05-09
Feature/text wrap

LukasFritzeDev:
Adds support for [CSS property `text-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap#values) to `Text` and `Heading` component. Use cases for this are described in the Issue https://github.com/mittwald/flow/issues/1437

The follow values are not allowed:
* `stable`: same behavior as wrap, except that when the user is editing the content. Since this scenario is not used in flow, it’s not required.
* `pretty`: allowed for `Text`, disallowed for `Heading`. This is intended for body copy by design, so not necessary for headings. 
* `nowrap`: disallowed in favour of a `Elipsis` component.

### DIFF
--- a/packages/components/src/components/Heading/Heading.module.scss
+++ b/packages/components/src/components/Heading/Heading.module.scss
@@ -69,6 +69,13 @@
   @include size("l");
   @include size("xl");
   @include size("xxl");
+
+  &:where(.wrap-wrap) {
+    text-wrap: wrap;
+  }
+  &:where(.wrap-balance) {
+    text-wrap: balance;
+  }
 }
 
 @media (max-width: 550px) {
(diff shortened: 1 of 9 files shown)


packages/components/src/components/Heading/Heading.module.scss
```diff
@@ -69,6 +69,16 @@
   @include size("l");
   @include size("xl");
   @include size("xxl");
+
+  &:where(.wrap-wrap) {
+    text-wrap: wrap;
+  }
+  &:where(.wrap-nowrap) {
+    text-wrap: nowrap;
+  }
```
LukasFritzeDev, line 78: Should `nowrap` be disallowed? It might break the layout.

> Text does not wrap across lines. It will overflow its containing element rather than breaking onto a new line.

LukasFritzeDev: Should be removed in favour of a `Elipsis` component.

Lisa18289 (approved):

PR #1465, opened 2025-05-15
fix: fix RHF form reset issue

mfal:


ins0:
Looks like this breaks stuff - the `defaultValue` prop from eg. `TextFields` will not be applied. Same case with RHF - the TextField will not reset to it's provided default value when a reset is performed with a native `<Button type="reset">`.

Lisa18289 (approved):

PR #1491, opened 2025-05-22
fix(List): fix pagination

Lisa18289:


packages/components/src/components/List/model/pagination/BatchesController.ts
```diff
@@ -24,6 +24,10 @@ export class BatchesController<T> {
   }
 
   public hasNextBatch(): boolean {
+    const filtertRowCount = this.reactTable.getFilteredRowModel().rows.length;
```
elmartinezc, line 27: `filteredRowCount`

elmartinezc (changes requested):

elmartinezc (approved):

PR #1499, opened 2025-05-23
fix(TextField): fix TextField loses value on blur

mfal:


packages/components/src/components/TextFieldBase/TextFieldBase.tsx
```diff
@@ -62,15 +62,22 @@ export const TextFieldBase: FC<TextFieldBaseProps> = (props) => {
     },
   );
 
+  /** Prevent weired reset behavior when value is 'undefined' */
```
Lisa18289, line 65: "weird" statt "weired"

Lisa18289 (approved):

PR #1587, opened 2025-06-10
Feature/simple form files

ins0:
- [x] simple and action forms will now behave like there native html elements and will receive a `FormData` object
- [x] remoteTunnel will now be auto generated
- [x] add remote form tests
- [x] adjust demo pages


packages/remote-core/src/serialization/FlowThreadSerialization.ts
```diff
@@ -27,6 +27,9 @@ export class FlowThreadSerialization extends ThreadSerializationStructuredClone
           if (isObjectType(val)) {
             return serialize({ ...val });
           }
+          if (isArray(val)) {
```
mfal, line 30: This is not required, because it is already handled by the base serializer (see `isSerializableByBase`).

mfal (changes requested):

PR #1599, opened 2025-06-12
feat(DonutChart): add donut chart component

Lisa18289:


packages/components/src/components/DonutChart/DonutChart.module.scss
```diff
@@ -0,0 +1,73 @@
+.donutChart {
+  position: relative;
+  width: fit-content;
+  flex-shrink: 0;
+
+  .border {
```
mfal, line 6: Actually you do not need to nest this classes (`border`, `background`, `value`, ...) because we have scoped class class names anyway.

packages/components/src/components/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,101 @@
...
+  size?: "m" | "l";
+  /** Divides the fill of the donut chart into segments */
+  segments?: DonutChartSegment[];
+  /** An optional text to be displayed instead of the formatted value */
+  valueText?: string;
+  /**
+   * The position of the legend, if set to false, the legend is hidden @default
+   * false
+   */
+  showLegend?: "top" | "left" | "bottom" | "right" | false;
```
mfal, line 32: This is my personal opinion, but I think it is confusing to mix the positioning and visibility. Someone not using TypeScript would be lost. Why not

```
  showLegend?: boolean;
  legendPosition?: "top" | "left" | "bottom" | "right";
  ```

File: packages/components/src/components/DonutChart/components/DonutChartLegend.tsx [hunk 3]
```diff
@@ -0,0 +1,28 @@
...
+  if (!segments || !showLegend) {
+    return null;
+  }
+
+  return (
+    <Legend className={styles.legend}>
+      {segments.map((s, i) => (
+        <LegendItem
+          color={s.color ?? getCategoricalColorByIndex(i)}
+        >{`${s.title} (${s.valueText ?? (formatOptions ? formatter.format(s.value) : `${s.value} %`)})`}</LegendItem>
```
mfal, line 24: This expression is hard to read/maintain. Can you split it up?

packages/components/src/components/DonutChart/components/DonutChartValue.tsx
```diff
@@ -0,0 +1,40 @@
...
+    !valueAndUnit ||
+    !valueAndUnit[1] ||
+    !valueAndUnit[2]
+  ) {
+    return (
+      <span className={styles.value}>
+        <b>{formattedValue}</b>
+      </span>
+    );
+  }
```
mfal, line 32: I think this component is doing to much implicit magic that could be better solved by allowing children here.

packages/components/src/components/DonutChart/DonutChart.module.scss
Lisa18289: i will leave it for now, as this is the case in all components, but we can think about refactoring them all

PR #1659, opened 2025-06-24
fix(Remote): rework postMessage serialization

ins0:
- [x] fixes serialisation for Transferable
- [x] mark arraybuffer directly as Transferable
- [x] add tests
- [x] add separated rhf form remote element

### DIFF
--- a/packages/remote-react-components/src/integrations/react-hook-form/Form/Form.tsx
+++ b/packages/remote-react-components/src/integrations/react-hook-form/Form/Form.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { type ComponentProps } from "react";
 import { Form as RhfForm } from "@mittwald/flow-react-components/react-hook-form";
 import type { FormProps } from "@mittwald/flow-react-components/react-hook-form";
 import type { FieldValues } from "react-hook-form";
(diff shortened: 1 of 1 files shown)


packages/remote-core/src/serialization/Serializer.ts
mfal, line 1: Except of the `export Key` change, I do not see any need for the other changes. Maybe a left-over from in-between-development?

packages/remote-core/src/serialization/FlowThreadSerialization.ts
```diff
@@ -24,10 +25,13 @@ export class FlowThreadSerialization extends ThreadSerializationStructuredClone
...
             const serialization = serializer.serialize(val);
             if (serialization.applied) {
-              return serialization.result;
+              return serialize(serialization.result);
             }
           }
           if (isObjectType(val)) {
+            if ((val as never)[TRANSFERABLE]) {
+              return serialize(val);
+            }
```
mfal, line 34: I would prefer to see this check in the `isSerializableByBase` function, because its actually a "just use the base serialization"-case. And maybe use `TRANSFERABLE in val` instead of `(val as never)[TRANSFERABLE]`.

```typescript
  private isSerializableByBase(val: unknown) {
    return (
      val instanceof Map ||
      val instanceof Set ||
      Array.isArray(val) ||
      typeof val === "function" ||
      (isObjectType(val) && TRANSFERABLE in val)
    );
  }
   ```

PR #1739, opened 2025-07-11
refactor(Autocomplete)!: use Option component to define autocomplete …

ins0:
…options


ins0:
- when in use on remote there are visual bugs
- using the component in the docs - the ui freezes

image

PR #1779, opened 2025-07-24
feat(ChartTooltip): remove tip from chart tooltip

Lisa18289:


packages/components/src/components/ChartTooltip/components/TooltipContent.tsx
```diff
@@ -1,7 +1,5 @@
 import clsx from "clsx";
-import styles from "../ChartTooltip.module.scss";
```
LukasFritzeDev, line 2: You did not remove all of the contents of this file, so it still contains some styles. Was the removal of the import here intentional?

Lisa18289: yes, the css is not used in this file anymore, but it is still used in the Tooltip.tsx file

PR #1900, opened 2025-08-19
docs: add target blank to external links

Lisa18289:


apps/docs/src/lib/mdx/components/MdxFileView/customComponents.tsx
```diff
@@ -89,19 +89,27 @@ export const customComponents = {
...
     if (href?.startsWith("http")) {
+      const url = new URL(href);
+
       return (
-        <Link href={href} inline>
+        <Link
+          href={href}
+          inline
+          target={
+            url.origin === "https://mittwald.github.io" ? undefined : "_blank"
```
ins0, line 99: origin is the origin url of the target url, this will always be the domain the app is running from. you need `hostname`

eg.
```suggestion
            url.hostname === "mittwald.github.io" ? undefined : "_blank"
```

PR #1987, opened 2025-09-16
add variant key to extension bridge config

freisenhauer:


Lisa18289 (approved):

mfal (changes requested):
- Please use this file https://github.com/mittwald/flow/blob/main/packages/ext-bridge/src/sessionToken/schemas.ts
- Please use conventional commit message

PR #2007, opened 2025-09-24
perf: use memo in provider components

mfal:


packages/components/src/lib/componentFactory/flowComponent.tsx
```diff
@@ -73,13 +75,15 @@ export function flowComponent<C extends FlowComponentName>(
       typeof ImplementationComponentType
     >;
 
-    const componentProps = useComponentPropsContext(componentName) ?? {};
+    const emptyProps = useMemo(() => ({}), []);
```
ins0, line 78: just define outside the component, no need to pass through `useMemo`

PR #2072, opened 2025-10-15
refact(Activity): use Activity from React if available

mfal:


packages/components/src/components/Activity/Activity.tsx
```diff
@@ -38,11 +59,15 @@ export const Activity: FC<ActivityProps> = (props) => {
     return isActive ? children : null;
   }
 
-  return (
-    <Suspense fallback={null}>
-      <SuspenseTrigger show={!isActive} />
-      {children}
-    </Suspense>
-  );
+  if (isActivitySupported && !forceCustomActivity) {
```
ins0: why the check and custom activity below at all? the components package has the dependency for react >= 19.2?

Issue #1130, opened 2025-01-24
Label hidden prop not working


danieldeichfuss:
### Project Information
mStudio Extension

### Feedback
We have a very simple form with only one input field. The extension is a note taking extension. We believe in this scenario showing the label would just visually clutter the UI. This is why we would like to visually hide the label without impacting the accessibility of this text area.

But the `hidden` prop of the [`<Label />`](https://mittwald.github.io/flow/03-components/content/label) component has no effect.

```
<Label hidden={true}>Notiz</Label>
```

<img width="952" alt="Image" src="https://github.com/user-attachments/assets/a528f9f8-78a8-4b50-8b0b-e5a49a4b352a" />

Lisa18289:
You can just remove the label and add a "aria-label" prop to the TextArea. But your right, the hidden prop should also have an effect, we will have a look at that issue

danieldeichfuss:
> You can just remove the label and add a "aria-label" prop to the TextArea. But your right, the hidden prop should also have an effect, we will have a look at that issue

Good idea. Thank you. Will do that for now.

PR #2087, opened 2025-10-21
fix(Label): support label hidden prop

ins0:
fixes https://github.com/mittwald/flow/issues/1130


mfal:
The `hidden` prop is just passed down to the underlying HTML-Element, what is fine in my opinion. If someone really needs to visually hide the Label, one could use `useVisuallyHidden` or something similar on the Label.

PR #2148, opened 2025-11-25
2131 extend the color component

Lisa18289:


packages/components/src/components/Color/Color.tsx
```diff
@@ -1,17 +1,38 @@
...
 /** @flr-generate all */
-
 export const Color: FC<ColorProps> = (props) => {
   const { children, color = "blue" } = props;
 
-  return <span className={color ? styles[color] : undefined}>{children}</span>;
+  const isFlowColor = flowColors.includes(color as FlowColor);
+
+  const className = isFlowColor ? styles[color as FlowColor] : undefined;
+  const style = !isFlowColor ? { color } : undefined;
```
mfal: Less type hacks

```suggestion
function isFlowColor(something: unknown): something is FlowColor {
  const anyFlowColors = flowColors as readonly string[];
  return typeof something === "string" && anyFlowColors.includes(something);
}

export interface ColorProps extends PropsWithChildren {
  /**
   * The color of the element.
   *
   * @default "blue"
   */
  color?: FlowColor | string;
}

/** @flr-generate all */
export const Color: FC<ColorProps> = (props) => {
  const { children, color = "blue" } = props;

  const isAFlowColor = isFlowColor(color);

  const className = isAFlowColor ? styles[color] : undefined;
  const style = !isAFlowColor ? { color } : undefined;
```

PR #2205, opened 2025-12-17
fix(Alert): add spacing between alerts

Lisa18289:


mfal (changes requested):
Components itself should not define an outer margin.

PR #2255, opened 2026-01-16
refactor(Action)!: drop deprecated action prop

mfal:
Use `onAction` instead


packages/components/src/components/Action/models/getExecutionFunction.ts
```diff
@@ -22,7 +22,7 @@ export function getExecutionFunction(action: ActionModel): ActionFn {
   };
 
   const {
-    action: actionProp,
+    onAction: actionProp,
```
Lisa18289: onAction ist doppelt

PR #2280, opened 2026-01-28
feat(Rating): make rating component interactive

Lisa18289:


packages/components/src/components/Rating/Rating.tsx
```diff
@@ -1,49 +1,93 @@
...
+  useMakeFocusable(localRef);
 
   return (
-    <div
-      aria-label={stringFormatter.format(`rating.${value}`)}
+    <Aria.RadioGroup
+      {...rest}
       className={rootClassName}
+      defaultValue={defaultValue.toString()}
+      value={value || value === 0 ? value.toString() : undefined}
```
mfal: `const stringValue = value?.toString();`

```suggestion
      value={stringValue}
```

packages/components/src/components/Rating/Rating.tsx
```diff
@@ -1,49 +1,99 @@
...
+          <FieldErrorCaptureContext>
+            <PropsContextProvider props={fieldPropsContext}>
+              {children}
+              <div className={styles.ratingSegments}>
+                {Array(5)
+                  .fill("")
+                  .map((_, index) => {
+                    const selectedValue = parseInt(
+                      renderProps.state.selectedValue ?? "0",
+                    );
```
mfal: I ment to put the assignemnt outside the loop, because it only depends on the props.

mfal (changes requested):

mfal (approved):

PR #2395, opened 2026-03-23
2076 add range mode to list filter

Lisa18289:


packages/components/src/components/List/model/filter/Filter.ts
```diff
@@ -162,6 +169,20 @@ export class Filter<T, TProp extends PropertyName<T>, TMatchValue> {
...
+    if (
+      value &&
+      typeof value === "object" &&
+      "start" in value &&
+      "end" in value
+    ) {
+      return value as DateRange;
+    }
+    return null;
+  }
```
mfal: Move this into own utility function (type guard) under `/components/Calendar/components/RangeCalendar/types.ts`

```ts
export function isDateRangeValue(value: unknown): value is DataRange {
 // ...
}
```

mfal (approved):

PR #2439, opened 2026-04-09
feat(Ext-Bridge): support language config in schema

mfal:


maaaathis (approved):

packages/ext-bridge/src/config/schemas.ts
```diff
@@ -6,6 +6,7 @@ const baseConfig = z.object({
   extensionId: z.string(),
   extensionInstanceId: z.string(),
   variantKey: z.string().optional().nullable(),
+  language: z.string().default("de-DE"),
```
martin-helmich: From what I've seen, the mStudio only sends a two-letter language code (meaning, either `de` or `en`), and not a locale string.


```suggestion
  language: z.string().default("de"),
```

ins0: This is correct, mStudio only support a general language identifier without a regional specifier - our flow intl supports both. from my point of view this is ok since it is a fallback mainly for flow

ins0 (approved):

PR #2440, opened 2026-04-09
feat(List): support custom empty views

mfal:


Lisa18289 (approved):

apps/docs/src/content/04-components/structure/list/overview.mdx
```diff
@@ -141,6 +141,19 @@ Die Pagination ist standardmäßig bei jeder List aktiviert, kann jedoch über
 `hidePagination` deaktiviert werden. Über die `batchSize`-Property kann
 festgelegt werden, wie viele Einträge gleichzeitig angezeigt werden sollen.
 
+## Empty View
+
+Mit `emptyView` kann eine benutzerdefinierte Ansicht angezeigt werden, wenn die
+Liste keine Einträge enthält. So lässt sich beispielsweise eine Illustration mit
+einem Hinweistext anzeigen, um die Nutzer:innen zu informieren, dass keine Daten
```
Lisa18289: das ":innen" muss weg, wir sollen nur gendern wenn es um z.B. Stellenausschreibungen geht

mfal:
We must have closer look into the layout-shifts from the initial loading state to the empty/non-empty list view.

Jan-Eimertenbrink:
> We must have closer look into the layout-shifts from the initial loading state to the empty/non-empty list view.

In general, I think it’s a good decision to integrate the initial EmptyView directly into the list. In the past, we often saw cases where the EmptyView for initial empty states where no items exist yet was simply forgotten by developers, so this approach helps avoid that. 

However, in the mStudio context, it does not feel visually appropriate to display a disabled ListHeader above the EmptyView. Because of this, I would prefer switching to a loading spinner during the initial loading state. A layout shift during loading cannot really be avoided anyway, regardless of whether we use a skeleton or a spinner. Using a spinner also keeps the option open to transition directly to an IllustratedMessage (EmptyView) without disabling the ListHeader. Visually, moving from fewer elements to more elements feels more natural and comfortable for users than the opposite direction.

Lisa18289 (approved):

Jan-Eimertenbrink (approved):

ins0 (approved):

PR #2444, opened 2026-04-10
Add light box gallery

### LINKED ISSUE #1836: Improve LightBox Component
## Description

There have already been two requests to extend the LightBox component with a gallery feature. This would allow users to click through multiple images within a given section.
We also already have a prominent use case for this in the Marketplace, where multiple visuals are often displayed together.

Additionally, we could explore ways to make it easier for contributors to embed videos or PDFs in the LightBox – for example, by providing suitable packages. However, this simplification will not be part of the current iteration.

## Design
We are extending the LightBox with a gallery feature.
To enable navigation between images, buttons will be displayed on the left and right edges of the viewport.
At the bottom of the viewport, we will add an indicator showing the user’s current position within the gallery and the total number of media items available.

Image

This indicator is purely visual for now, and we’ve decided to use dots, as they take up minimal space and keep the focus on the main content.
We will introduce a new design token, `lightbox-content-color`, using the color `light-800`. And a token for the indicator spacing `lightbox-indicator-spacing` with `size-xs`. For the icons, we use the small size.

Lisa18289:


Jan-Eimertenbrink (changes requested):
This is already feeling really good and close to what I had in mind. I have a few suggestions that could improve the experience further:

1. Keyboard navigation in the gallery
Once inside the gallery, I would expect to be able to navigate between images using the arrow keys. That would make keyboard interaction much smoother and aligns with common patterns, for example in the Google Play Store.

2. Spacing from the screen edges
I think the Buttons could have a bit more space from the border, especially on larger screens where they currently feel slightly too close to the borders. Maybe something around 32px could work, but we can review that together. I’d still keep them visually anchored toward the edges, similar to how it’s handled in the Google Play Store.

3. Gallery looping behavior (open for discussion)
I’m a bit undecided here and would love your thoughts. Should the gallery loop from the last image back to the first?

There are pros and cons:

- Looping makes it easy to continuously browse in one direction and quickly jump from the first to the last image.
- On the other hand, having a clear start and end (no looping) gives users a stronger sense of completion when they reach the last image.

I can see good arguments for both approaches, so I’d be interested in your perspective.

mfal (approved):

Jan-Eimertenbrink (approved):
Looks good now

PR #2447, opened 2026-04-13
feat(Remote): auto-set host language on remote-side

mfal:


packages/remote-react-components/src/components/RemoteRoot.tsx
```diff
@@ -154,7 +161,9 @@ export const RemoteRoot: FC<RemoteRootProps> = (props) => {
           >
             <Suspense fallback={loadingFallback}>
               <ViewComponentContextProvider components={viewComponents}>
-                {children}
+                <IntlProvider locale={hostConfigRef.current?.language}>
```
ins0: should be above since the suspense is not covered with the `IntlProvider`

ins0 (approved):

PR #2459, opened 2026-04-16
feat(CodeBlock): add expandable option

### LINKED ISSUE #2374: collapse code in messages
It would be great if longer code/output blocks could be collapsible (similar to how some other Markdown editors support this via a `<details>` block).

Lisa18289:


packages/components/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -12,6 +15,13 @@ export interface CodeBlockProps
       Pick<CodeEditorProps, "language" | "showLineNumbers" | "copyable">
     > {
   code?: string;
+  /**
+   * Enables collapsing of long code blocks with a "Show more" toggle. @default:
+   * false
+   */
+  expandable?: boolean;
+  /** Line count threshold before the code block is collapsed. @default: 10 */
+  expandAfterLines?: number;
```
ins0: We should think about the prop name. the complete feature is more a `truncateLines` with `boolean | number`. I think that better describes the property/function and what it actually does.

packages/components/src/components/CodeBlock/locales/de-DE.locale.json
```diff
@@ -0,0 +1,4 @@
+{
+  "codeBlock.showLess": "Weniger anzeigen",
```
ins0: we don't need a component name prefix anymore in the translation. if you prefer to prefix it - use the function context within the component e.g. `expand.showLess` or `button.showLess`

packages/components/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -32,16 +32,15 @@ export const CodeBlock: FC<CodeBlockProps> = (props) => {
...
-    expandAfterLines = 8,
+    truncateLines = false,
     ...rest
   } = props;
 
-  const [expanded, setExpanded] = useState(false);
+  const [collapsed, setCollapsed] = useState(true);
   const [maxHeight, setMaxHeight] = useState<number>();
-  const [shouldExpand, setShouldExpand] = useState(false);
+  const [shouldCollapse, setShouldCollapse] = useState(false);
```
ins0: i'm not sure why this state exists in the first place. `collapsed` and `maxHeight` should be enough. just don't set `setMaxHeight` when `totalLines > visibleLines` - no?

ins0 (approved):

PR #2466, opened 2026-04-21
feat(Tunnel): support Tunnel nesting

mfal:


packages/components/src/components/UiComponentTunnel/lib.ts
```diff
@@ -0,0 +1,6 @@
+import type { FlowComponentName } from "../propTypes";
+
+const tunnelProviderPrefix = "@mittwald/flow/";
```
ins0: we should make it clear in the prefix that this is a tunnel e.g. `@mittwald/flow/tunnel/`

packages/components/src/components/Heading/Heading.tsx
```diff
@@ -1,14 +1,14 @@
...
 import clsx from "clsx";
 import type { PropsContext } from "@/lib/propsContext";
 import { PropsContextProvider } from "@/lib/propsContext";
 import type { FlowComponentProps } from "@/lib/componentFactory/flowComponent";
 import { flowComponent } from "@/lib/componentFactory/flowComponent";
 import * as Aria from "react-aria-components";
-import { TunnelExit, TunnelProvider } from "@mittwald/react-tunnel";
+import { UiComponentTunnelExit } from "../UiComponentTunnel/UiComponentTunnelExit";
+
+export * from "./view";
```
ins0: this should be moved in the `index.ts`

packages/components/src/components/TextField/config.ts
ins0, line 1: left over?

ins0 (changes requested):

packages/components/src/components/UiComponentTunnel/lib.ts
mfal: This prefix is only used in the context of "Tunnel". It is not stored or used anywhere else, where it is "visible" from the outside or might be conflicting with other keys (localstorage, cookie, ...). So I don't think this is required.

ins0: i know - but i think it is good practice to give a prefix some information where its origin comes from.

ins0 (approved):

PR #2473, opened 2026-04-28
fix(Navigation): fix visual style

ins0:


packages/components/src/components/Navigation/Navigation.tsx
```diff
@@ -50,9 +51,17 @@ export const Navigation = flowComponent("Navigation", (props) => {
...
+          children={(children) => {
+            if (Children.count(children) >= 1) {
+              return <ul>{children}</ul>;
+            }
+
+            return null;
+          }}
+          id="links"
+          component="Navigation"
+        />
```
mfal: Can you add a sub-component for conditionally rendering the links list and use it here and in the `<NavigationGroup/>`? Except of the `tunnelId` it is always the same code.

mfal (approved):

PR #2479, opened 2026-05-04
fix(CodeBlock): fix scrollbar position

### LINKED ISSUE #2476: Horizontal scrollbar overlaps content in single-line code blocks
A user reported a problem with code blocks on Windows.

When a code block contains only a single line and is too long, it requires horizontal scrolling. On Windows, the horizontal scrollbar can overlap with the content. This makes the text harder to read and interact with.

https://github.com/user-attachments/assets/e12e23d8-6473-408e-94ae-687b63644024

This is filmed on Mac because i cant show you the original.

This seems to be related to how scrollbars are displayed on Windows. Instead of overlaying the content, they take up space inside the layout. In very short code blocks (for example, one line), the scrollbar appears in the same vertical space as the text and can partially cover it. On macOS, a similar issue can appear, but it is usually less noticeable. Scrollbars often overlay the content and disappear when not in use. Users can also change system settings to always show scrollbars, which places them below the content and avoids the overlap.

This behavior seems to be influenced by the operating system. Is there something we should fix on our side (for example layout or spacing adjustments), or is this something we should accept as a platform-specific behavior?

Lisa18289:


mfal (changes requested):
Could you provide a visual test where the scrollbar is actually visible?

mfal (approved):

PR #2517, opened 2026-05-28
Popover datepicker optimization

mfal:


Lisa18289:
it does not look good without padding,  maybe we should add a --calendar--cell-size-mobile instead?

Lisa18289:
we need this optimization for date range picker as well

packages/components/src/components/DatePicker/DatePicker.tsx
```diff
@@ -32,6 +33,11 @@ export const DatePicker = flowComponent("DatePicker", (props) => {
 
   const rootClassName = clsx(fieldProps.className, className);
 
+  const calendarClassName = clsx(
+    styles.calendar,
+    getBodyInnerWidth() < 300 && styles.small,
```
Lisa18289: could  we not  just use css media query instead?

mfal:
Reducing the padding is not desired.

PR #2541, opened 2026-06-10
feat(Markdown): allow custom components to override default renderers

TimAufdemKampe:


mfal:
I'm having a look soon.

packages/components/src/components/Markdown/Markdown.tsx
mfal: Weil das via Remote vermutlich nicht gehene wird.

```suggestion
/** @flr-generate all
      @flr-ignore-props components
  */
```

mfal (changes requested):

mfal (approved):

PR #2569, opened 2026-06-22
test: fix flaky browser test

mfal:


packages/components/src/components/Modal/Modal.browser.test.tsx
```diff
@@ -267,10 +269,14 @@ test("useOnClosed is called when the closing animation has finished", async () =
   expect(onClosed).not.toHaveBeenCalled();
 
   await userEvent.click(closeButton);
+  await sleep(50);
```
ins0, line 272: looks promising to skip real sleeps https://vitest.dev/api/vi.html#vi-advancetimerstonextframe

ins0 (approved):

PR #2574, opened 2026-06-24
fix(NavigationGroup): remove unnecessary labelled by

Lisa18289:


mfal:
We still need this for NavigationGroups with a Label. Maybe only set `aria-labeledby` if `aria-label` does *not* exists.

mfal (approved):

PR #2596, opened 2026-07-02
feat(Remote): add async serializers

ins0:


apps/remote-dom-demo/src/app/remote/list/page.tsx
```diff
@@ -2,6 +2,7 @@
 
 import { demoData } from "@/app/remote/list/demoData";
 import {
+  Action,
```
mfal, line 5: Is this import used somewhere?

patches/@quilted__threads@3.3.1.patch
mfal, line 1: I've seen that a new major of  Quilted Threads was released. Maybe this supports async serialization already?

packages/remote-react-renderer/src/components/lib/prepareFormData.test.ts
mfal, line 1: Can you leave a short comment why this test is dropped?

mfal (changes requested):

apps/remote-dom-demo/src/app/remote/list/page.tsx
ins0: Yes in the file, its currently broken.

patches/@quilted__threads@3.3.1.patch
ins0: Nope, not supported. If we consider to upgrade in the future to v4, we can try to provide a PR to quilted threads with these changes.

packages/remote-react-renderer/src/components/lib/prepareFormData.test.ts
ins0: it's not required anymore to prepare the formData. The function `prepareFormData` that was tested in this test is deleted.

mfal (approved):

PR #2605, opened 2026-07-06
refactor(ContextualHelp): simplify structure and add contextual trigger label

### LINKED ISSUE #2504: Contextual Help uses unnecessary heading structure
The Contextual Help popover currently uses an H5 heading, followed by text and additional information. This adds unnecessary semantic structure, since the content is usually a single, simple explanation.

A heading is not required here and makes the structure more complex than needed. A bold text label (or even no heading at all) would be sufficient to visually indicate structure.

In addition, the content should not be split into multiple paragraphs. Contextual help should stay compact and focused, without breaking the information into sections.

We should also ensure that the trigger uses a clear, descriptive aria-label. Instead of a generic label like “More information”, it should describe the context, for example: “More information about XY”.

maaaathis:
closes #2504

### DIFF
--- a/apps/docs/src/content/04-components/overlays/contextual-help/guidelines.mdx
+++ b/apps/docs/src/content/04-components/overlays/contextual-help/guidelines.mdx
@@ -11,6 +11,16 @@ Achte bei der Verwendung des ContextualHelp darauf, dass ...
 - es nicht für Inhalte benutzt wird, die zum Verständnis der Benutzeroberfläche
   unerlässlich sind. Stattdessen wird es für weiterführende Erklärungen oder zum
   Verweisen auf externe Quellen verwendet.
+- der Inhalt kompakt bleibt und nicht in mehrere Absätze oder Abschnitte
+  aufgeteilt wird. Eine Überschrift ist nicht notwendig.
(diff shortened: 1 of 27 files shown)


apps/docs/src/content/04-components/overlays/contextual-help/guidelines.mdx
```diff
@@ -11,6 +11,10 @@ Achte bei der Verwendung des ContextualHelp darauf, dass ...
 - es nicht für Inhalte benutzt wird, die zum Verständnis der Benutzeroberfläche
   unerlässlich sind. Stattdessen wird es für weiterführende Erklärungen oder zum
   Verweisen auf externe Quellen verwendet.
+- der Inhalt kompakt bleibt und nicht in mehrere Absätze oder Abschnitte
+  aufgeteilt wird. Eine Überschrift ist nicht notwendig.
+- über die `label` Property des `<ContextualHelpTrigger />` beschrieben wird,
```
mfal: It should also be mentioned, that the label is more used like a "subject" which is used in the internationalized "label-template" `More information about {subject/label}`.

mfal (changes requested):

mfal (approved):

PR #2616, opened 2026-07-07
docs: refactor documentation of for multi upload

### LINKED ISSUE #2371: Add Code Snippet for Multiple Upload
## Description

With the CronSyntax issue we introduced a new category called Code Snippets. The idea behind this category is to document small UI patterns that are not large enough to be considered full patterns, but still contain reusable custom code that appears in multiple places.

The topic of multiple uploads fits well into this category. At the moment, the guidance and code are described in the FileField and Dropzone documentation within the “Kombiniere mit…” section. Since the underlying approach can be reused independently of these specific components, it could make sense to move this into a dedicated Code Snippet.

Before creating it, we should briefly align on what exactly the snippet should cover and which aspects we want to describe.

Lisa18289:


apps/docs/src/content/03-patterns/02-codesnippets/multi-upload/examples/default.tsx
```diff
@@ -13,19 +15,20 @@ export default () => {
 
   return (
     <Section>
-      <FileField
+      <FileDropZone
         multiple
         onChange={(value) => {
           if (value) {
             setFiles([...files, ...value]);
```
ins0:
```suggestion
        onChange={(droppedFile) => {
          if (value) {
            setFiles((existingFiles) => [...existingFiles, ...droppedFile]);
```

ins0 (changes requested):

ins0 (approved):

PR #2634, opened 2026-07-10
feat(TabNavigation): introduce tab navigation

### LINKED ISSUE #1877: Revisit Responsive Tabs Behavior
## Problem

I noticed an issue that will come up when we implement the new Flow Library in mStudio. The original Tabs component was not designed to handle a very large number of tabs. The context has changed since extensions can now add their own tabs.  

If users install many extensions or if tab names are particularly long, we reach the limit even on large viewports, causing tabs to move into a context menu. This behavior was intentionally implemented in the past due to previous issues (more details in this [issue](https://github.com/mittwald/flow/issues/336)). Overall, it worked well for five to six tabs and in mobile or tablet contexts, where the menu and the currently selected tab are close together.  

However, since theoretically extensions can now add unlimited tabs, the responsive behavior is triggered even on the desktop version. This causes the context menu for tab selection to appear far from the currently selected tab (see image), which I believe reduces its visibility and focus for the user.  

Image

## Possible Solutions (to-discuss)

There are several ways we could address this. We could display the context menu on the left (Like [Spectrum](https://spectrum.adobe.com/page/tabs/#Tab-overflow) does)to align it more with reading flow. Another option would be to consider horizontal swipeable tabs again (see video). This would need further review, as swipe handling in the old Flow was not ideal, and I am not sure about its accessibility.  

https://github.com/user-attachments/assets/8d656421-61c6-401c-a84e-1cce2fd144e5

We can use this issue to gather ideas so that when we rebuild mStudio, we can potentially address this usability problem with the tabs upfront.

### LINKED ISSUE #2680: Introduce TabNavigation component and repurpose Tabs for content switching
Refactor the existing tab components to clearly separate content switching from navigation.

**Tasks**
- Restyle the existing Tabs component while keeping its functionality unchanged.
- Use Tabs exclusively for switching between different content panels.
- Introduce a new TabNavigation component for navigation purposes.
- TabNavigation should retain the visual appearance of the current Tabs component.
- TabNavigation is purely a navigation component and should not be used for content switching.

**Goal**
Separate the concepts of content tabs and navigation tabs to make their intended use clearer and provide dedicated components for each use case.

Lisa18289:


mfal:
Note for when this lands — per **ADR 0003** (component lifecycle status) and the 1.0 status curation in #2711, the new `TabNavigation` component should be tagged **`new`**. It's on Jan's "New" list but couldn't be included in the initial status-annotation PR because the component doesn't exist yet.

Two steps when picking this back up:
1. **Update this branch against `main`** — it already carries the ADR 0003 status-registry infrastructure (#2729).
2. Add `@flowStatus new` to the component-level JSDoc (alongside `@flr-generate` if it's remote-capable), e.g.:

```tsx
/**
 * @flr-generate all
 * @flowStatus new
 */
export const TabNavigation = flowComponent("TabNavigation", (props) => {
```

Then regenerate with `pnpm nx build:status-registry components` and commit `src/status/component-status.json`.

packages/components/src/lib/hooks/dom/useObserveOverflow.ts
```diff
@@ -0,0 +1,92 @@
+import type { RefCallback } from "react";
```
mfal, line 1: Why is this moved from `packages/components/src/lib/hooks/dom/useObserveOverflow.tsx` to here? Isn't it a general hook?

packages/components/src/components/TabNavigation/TabNavigation.module.scss
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/* Menu items of collapsed links (rendered inside the context menu popover,
+   outside of the navigation element). */
+.menuItem {
+  &.current,
+  &.current:hover {
+    background-color: var(--menu-item--background-color--current);
+    font-weight: var(--menu-item--font-weight--current);
+  }
```
mfal: This styling should already be applied by the ContextMenu > MenuItem itself, iguess.

packages/components/src/components/Tabs/components/TabTitle/TabTitle.module.scss
```diff
@@ -28,20 +32,20 @@
...
-    background-color: var(--tabs--background-color--default);
+    color: var(--tabs--color--disabled);
   }
 
   .text {
+    position: relative;
     display: flex;
     align-items: center;
     column-gap: var(--tabs--spacing);
+    z-index: 2;
```
mfal: I would like to see no fixed z-index anywhere, because this potentially may result in strange overlapping glitches with other overlays.

apps/docs/src/content/04-components/structure/tabs/overview.mdx
mfal, line 1: Tabs is more of "Structure" category I think.

packages/components/src/lib/dom/createFrameResizeObserver.ts
mfal, line 1: Tests missing

mfal (changes requested):

packages/components/src/lib/hooks/dom/useObserveOverflow.ts
Lisa18289: You are right — it is a general hook and nothing in it is Tabs-specific. Moved back to `packages/components/src/lib/hooks/dom/useObserveOverflow.ts` and the comment about children being portaled in is now phrased generically (a `UiComponentTunnel` is just one reason why the observing component never sees a `children` prop).

packages/components/src/components/TabNavigation/TabNavigation.module.scss
Lisa18289: Good catch, and it should indeed be the `MenuItem`. It could not be, though: React Aria's `filterDOMProps` only lets `id`, a few labeling/link props, global attributes and `data-*` through — `aria-current` is dropped before it reaches the DOM, so the `menuItem` mixin's `[aria-current]` rule never matched for a `MenuItem` (verified in the browser: the rendered `<a>` had no such attribute).

`MenuItem` now owns the state: it accepts `aria-current` and mirrors it to `data-current`, which survives the filter, and the shared `menuItem` mixin matches `[data-current]` next to `[aria-current]` (the latter is still needed for `Navigation`, whose items are real anchors). The duplicated block in `TabNavigation.module.scss` is gone and `LinkContent` just passes the prop on.

Covered by a new browser test in `TabNavigation.browser.test.tsx`; the `TabNavigation-context-menu-opened` screenshot renders byte-identically to before.

packages/components/src/components/Tabs/components/TabTitle/TabTitle.module.scss
Lisa18289: Removed — there is no fixed z-index in Tabs anymore (and the `isolation: isolate` that guarded it is gone too).

The active indicator precedes `.titles` in the DOM, so it is painted below the titles by tree order alone. The only thing the z-index was actually buying was that the selected tab's hover/pressed background did not cover the indicator — that is now expressed directly: hover/pressed backgrounds only apply to `:not([data-selected])`, since the selected tab already has the indicator as its background.

Verified as a rendering no-op: the `Tabs states`, `Tabs interaction` and `Tabs edge cases` screenshots come out byte-identical to before the change. One intentional difference remains, which is not covered by a screenshot: while the indicator slides to a newly selected tab, that tab no longer shows a hover background underneath it.

apps/docs/src/content/04-components/structure/tabs/overview.mdx
Lisa18289: Agreed, moved: `04-components/navigation/tabs` → `04-components/structure/tabs`, all incoming links updated (tab-navigation and counter-badge guidelines, section guidelines, layout and Informationskonzept foundations), and the Storybook title is now `Structure/Tabs`. Note this changes the public docs URL — the site is a static export without redirects, so old links to `/04-components/navigation/tabs/…` will 404.

packages/components/src/lib/dom/createFrameResizeObserver.ts
Lisa18289: Added `createFrameResizeObserver.browser.test.ts` — a browser test rather than a unit test, so it runs against a real `ResizeObserver`/`requestAnimationFrame` instead of mocks. It covers the deferral to the next frame, coalescing of repeated `schedule()` calls, measuring on an observed resize, `observe()` dropping the previous targets, and `disconnect()` stopping both further notifications and an already queued measurement.

packages/components/src/components/TabNavigation/components/LinkContent/LinkContent.tsx
```diff
@@ -0,0 +1,62 @@
...
+    download,
+    hrefLang,
+    ping,
+    referrerPolicy,
+    routerOptions,
+    "aria-label": ariaLabel,
+    "aria-current": ariaCurrent,
+  } = linkProps;
+
+  const isCurrent = ariaCurrent === "page";
```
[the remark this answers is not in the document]
Lisa18289: Fixed. `isCurrent` is gone entirely — `LinkContent` now forwards `aria-current` to the `MenuItem`, which treats every value except an explicit `"false"` as current, matching the `[aria-current]:not([aria-current="false"])` selector in `useCollapsingItems`.

mfal (approved):

PR #2653, opened 2026-07-15
chore(remote-dom-demo): unify demo content into a Star Wars theme

mfal:
## What

Rethemes **every** page of the `remote-dom-demo` app around a single, coherent **Star Wars** context, entirely in **English**.

Previously the demo pages mixed German and English, lorem ipsum, and generic placeholders (`Foo`/`Bar`/`Baz`, `Press me`, `Bla`, a random international person/language roster), so they read as unrelated snippets rather than one app.

## Why

The demo is what extension developers look at to understand the remote-rendering stack. Consistent, real-feeling content makes it far easier to follow — and a single running theme makes the examples feel like one product instead of scattered test fixtures.

## Key changes

- **`list/demoData.ts`**: replaced the ~250 lorem-ipsum entries with a small deterministic generator that produces ~250 Star Wars characters (named heroes + generated clone/droid/pilot designations) with `faction`, `homeworld`, and in-universe bios. IDs/filter values are stable across renders (no `Math.random`). The list filter field was renamed `language` → `faction`.
- **Locale**: `host/layout.tsx` `IntlProvider` switched `de-DE` → `en-US`, so the Flow components' built-in strings (`List`, `Search`, select placeholders, date pickers, …) render in English. The `TranslationProvider` demo in `react-hook-form` was updated to keep working under `en-US`.
- **Content sweep** across all pages: form labels/options (Clearance, Rank, Homeworld, Holocrons, Squadron), modal bodies (squadron creation, "Destroy the Death Star"), chart series (Shields/Hull), notifications, navigation, suspense, alerts, and error/console strings.

## Notes for reviewers

- **Content-only change** — no component behaviour, design tokens, or generated artifacts are touched.
- Mechanism/API-demonstration labels are intentionally left generic (`SubmitButton`, `SubmitController`, `Native Reset`, and the modal variant names `Offcanvas`/`With columns`/`Lightbox`/`With controller`) since they name the feature being demonstrated.
- Faction values cycle by index over the character list, so some pairings are deliberately non-canonical (e.g. "Leia Organa (Galactic Empire)") — this is intentional variety for the filter/demo, not a data bug.
- Verified in-browser (List, forms, chart, modal, navigation, event-handler) plus `tsc --noEmit`, ESLint, and Prettier all pass for the app.


maaaathis (approved):
glhf

PR #2671, opened 2026-07-20
fix(Modal): keep mobile modals with forms usable when the virtual keyboard opens (#2423)

### LINKED ISSUE #2423: Modals with forms hard to use on mobile
### Flow Version

_No response_

### Bug Description

On mobile devices modals (especially one with forms) are very hard to use, because the modal content is to small. The situation gets critical when the virtual keyboard pops up, i.e. when clicking into a textfield. See the screenshots (orange is the scrollable content).

Image

After clicking in the textfield:

Image

### Steps to Reproduce

1. Go to https://mittwald.github.io/flow/04-components/overlays/modal/overview on your mobile
2. Open the modal
3. Click into the text field

### Expected Behavior

1. The focused textfield is still in the viewport
2. Header and action buttons are not fixed, so there is more space to edit the form. If there are three action buttons, there is less then 100px space for the modal content on small devices.

### Actual Behavior

1. The focused textfield is not visible
2. The modal content is very small that makes filling out the form very hard.

### How much does this bug impact you?

🟠 Major – It breaks important functionality

### Relevant Logs or Error Output

```shell

```

### Browser Version

_No response_

### Before Submitting

- [x] I have checked that this issue has not already been reported.
- [x] I have provided all necessary information to reproduce the issue.

mfal:
## Problem

Fixes #2423. On mobile, a Modal with a form was cramped: the fixed header and fixed action footer squeezed the scrollable content to well under 100px once the on-screen keyboard opened, and the focused field could scroll out of view.

## Solution

On mobile (`@media (max-width: 768px)`, non-off-canvas) the Modal is now a **single scroll container**:

- **At rest** the header is `sticky; top: 0` and the footer is `sticky; bottom: 0` — the familiar docked look, but in one scroll region (the dialog scrolls).
- **While the virtual keyboard is open** both un-stick (`position: static`) so header + content + footer flow normally and the content gets the full available space. The dialog fills the available height and the footer uses `margin-top: auto`, so with **short** content the footer still sits at the bottom (not mid-air) and with **tall** content it scrolls away with everything else. The footer is never hidden, so the primary action stays reachable.

Keyboard state is detected with a small `useVirtualKeyboardVisible` hook via the VisualViewport API (`window.innerHeight - visualViewport.height > 150`), which toggles a `keyboardVisible` class. This builds on the existing react-aria `--visual-viewport-height` placement (#2332/#2301) rather than reverting it.

Desktop and off-canvas behavior are unchanged.

Also included: a **behavior-neutral cleanup** of `Modal.module.scss` + `Overlay.module.scss` (merged the two duplicate mobile media queries, section headers, documented the react-aria CSS variables).

## Key changes

- `packages/components/src/lib/hooks/dom/useVirtualKeyboardVisible.ts` (+ unit test)
- `packages/components/src/components/Modal/Modal.tsx` — consumes the hook, toggles `keyboardVisible`
- `packages/components/src/components/Modal/Modal.module.scss` — mobile single-scroll + keyboard-aware sticky chrome; section cleanup
- `packages/components/src/components/Overlay/Overlay.module.scss` — comments/cleanup only
- `packages/components/src/components/Modal/Modal.browser.test.tsx` — asserts the mobile rest-state layout

## Testing

- Unit: `useVirtualKeyboardVisible` predicate (threshold boundaries)
- Browser: mobile Modal is a single scroll container with sticky header + footer (via `page.viewport(375, 500)`)
- `affected:test` green (unit + compile); Modal + Overlay browser suites green
- Verified on a real iOS simulator: header/footer un-stick while typing, footer reachable by scrolling, re-dock on blur; short-form footer stays at the bottom

## Notes for reviewers

- No public API change (no new Modal prop) → no remote/flr regeneration, no `doc-properties` change.
- CSS changes are entirely inside `@media (max-width: 768px)`; the visual snapshots run at desktop width, so **no snapshot changes are expected**. Please apply the `run-visual-tests` label to confirm (and `update-screenshots` only if something unexpectedly shifts).
- The virtual-keyboard detection uses the VisualViewport API; the `> 150px` threshold may want per-platform tuning if edge cases surface.


packages/components/src/components/Modal/Modal.module.scss
```diff
@@ -187,46 +205,104 @@
...
+   One scroll container: the dialog scrolls while the header sticks to the top
+   and the footer to the bottom. While the virtual keyboard is open
+   (.keyboardVisible) both un-stick so header + content + footer flow normally
+   and the content gets the full available space.
+   =========================================================================== */
 @media (max-width: 768px) {
   .modal {
-    > div {
-      > div {
-        @include scrollbar.scrollbar;
```
Lisa18289, line 196: die scrollbar ist wahrscheinlich beim mergen rausgeflogen, sollte aber eigentlich drin bleiben

Lisa18289 (approved):

PR #2685, opened 2026-07-22
fix(CodeBlock): fix undefined max height in code block

Lisa18289:


packages/components/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -58,7 +58,11 @@ export const CodeBlock: FC<CodeBlockProps> = (props) => {
   return (
     <div
       className={clsx(rootClassName, folded ? styles.folded : undefined)}
-      style={{ "--max-height": `${maxHeight}px` } as React.CSSProperties}
+      style={
+        maxHeight
+          ? ({ "--max-height": `${maxHeight}px` } as React.CSSProperties)
+          : undefined
+      }
```
mfal: Put this into a separate const to make the JSX part more readable.

mfal (approved):

PR #2688, opened 2026-07-23
fix(remote-react-components): scope cross-version refs per test file

mfal:
## Problem

The scheduled **Run scheduled visual tests** workflow's `cross-version` job has failed on every run since it was introduced (2026-07-21) — specifically the **in-process** harness, with a `Checkbox edge cases(CrossVersion)` structural-snapshot mismatch against the previous published version.

It is **not** a version incompatibility. The in-process harness keyed each ephemeral reference HTML file only on `testName + description`:

```ts
`${REF_DIR}/${slugify(`${testName}-${description}`)}.html`
```

Two reused visual tests share the exact name `"Checkbox edge cases"`:

- `Checkbox.browser.test.tsx` → renders one `<Checkbox>` (7-tag tree)
- `CheckboxButton.browser.test.tsx` → renders a `Flex` with two `<CheckboxButton>`s (18-tag tree) — the test is mis-named `"Checkbox edge cases"`

Both therefore read and write the **same** `.refs/*.html` file. Whichever test writes the reference last makes the other's comparison mismatch. This is deterministic (it failed on the job's very first scheduled run and every one since), and unrelated to the actual old version — current and the compared version render identically once the tests are keyed apart.

The normal visual screenshot suite does not hit this because vitest scopes screenshots per test file (`__screenshots__/<file>/…`); only the cross-version harness used a flat, non-file-scoped key.

## Fix

Include the source-file basename in the reference key, mirroring vitest's own per-file screenshot scoping. Small, self-contained change to `crossVersionEnvironment.tsx`. This resolves the collision and prevents any future same-name collision.

## Validation

- Installed `@mittwald/flow-remote-react-components@0.2.0-alpha.932` locally and reproduced the failure via the full 83-file corpus.
- After the fix: full-corpus comparison vs alpha.932 → **167/167 pass, run twice, deterministically green**; the reference pass now writes two distinct edge-case files instead of one.
- `test:compile` (tsc), ESLint, and Prettier all clean.
- No impact on the visual suite or committed screenshots — the `.refs` are wiped and regenerated on every run.

## Notes for reviewers

- Supersedes the approach in #2683, which bumps the cross-version skip watermark. That treats this harness bug as a version break (and targets `alpha.933`, a docs-only release that renders identically to `932`), so it would not durably fix the failure.
- **Optional follow-up:** rename the `CheckboxButton` test from `"Checkbox edge cases"` to `"CheckboxButton edge cases"` for clarity. Left out here because the description also names that test's committed visual screenshot baselines (`CheckboxButton.browser.test.tsx/Checkbox-edge-cases-*.png`), so renaming would require regenerating those (including the Linux baseline via the `update-screenshots` label). The file-scoping fix makes the duplicate name harmless regardless.
- **Separate, pre-existing flake (out of scope):** the *iframe* step's `Button` scenarios intermittently hit ~180–240s `@quilted/threads` connection-timeout hangs; the workflow's retry loop usually absorbs it.


packages/remote-react-components/e2e/cross-version-inprocess/crossVersionEnvironment.tsx
```diff
@@ -24,8 +24,27 @@ const slugify = (value: string): string =>
     .replace(/^-|-$/g, "")
     .toLowerCase();
 
-const referencePathFor = (testName: string, description: string): string =>
-  `${__CROSS_VERSION_REF_DIR__}/${slugify(`${testName}-${description}`)}.html`;
+// The reference file is keyed by the SOURCE FILE too, not just the test name +
+// description. Two reused tests in different files can share a name — e.g. both
+// Checkbox and CheckboxButton have a `Checkbox edge cases` test — which without
```
Lisa18289, line 29: ich glaub das sollte eigentlich nicht so sein, beim CheckboxButton sollte es auch "CheckboxButton edge cases" heißen

Lisa18289: achso, hab grad gesehen der test wurde weiter unten auch angepasst, aber dann wär die regel ja eher das test Namen immer als erstes Wort den Component Namen haben müssen oder?

mfal: Ja, genau — deshalb habe ich den Test weiter unten in `CheckboxButton.browser.test.tsx` auf `CheckboxButton edge cases` umbenannt (Commit 0f9003b). Die vier Screenshot-Baselines wurden per `git mv` mitgezogen (identischer Render, keine Neugenerierung nötig).

mfal: Genau, der Test heißt jetzt `CheckboxButton edge cases`. Die Konvention „Testname beginnt mit dem Komponentennamen" ist sinnvoll und der Rename folgt ihr.

Das Datei-Scoping hier (Zeile 29) habe ich trotzdem bewusst drin gelassen: eine reine Namenskonvention ist nicht erzwungen, und zwei verschiedene Testdateien könnten legitim gleichnamige Tests haben. Das Datei-Scoping macht Kollisionen strukturell unmöglich — so wie Vitest seine Screenshots ohnehin pro Testdatei ablegt. Die beiden Ebenen ergänzen sich also.

Falls die Konvention zusätzlich erzwungen werden soll, kann ich das gern als Follow-up per ESLint-Regel nachziehen.

Lisa18289 (approved):

PR #2703, opened 2026-07-23
chore(lint): enforce recurring PR-review rules via tooling

mfal:
## Why

Enforces, via tooling, several recurring issues that were being caught by hand
in PR reviews (an analysis of ~340 inline review comments). This follows Flow's
guiding principle §8: *enforce consistency with tooling, not discipline*.

Rule-enable commits are kept separate from the resulting-fix commits so each
side is easy to review.

## What

**ESLint**
| Rule | Level | Result |
|------|-------|--------|
| `react/jsx-key` | error | plugin was missing; 3 violations fixed |
| `react-hooks/exhaustive-deps` | warn | non-blocking; surfaces useEffect dep issues (72 warnings) |
| `@typescript-eslint/no-non-null-assertion` | error | off for generated code via the existing override; 5 violations fixed |

**Stylelint**
| Rule | Level | Result |
|------|-------|--------|
| `logical-css/require-logical-properties` + `require-logical-keywords` | error (autofix) | inline-direction only; 93 declarations converted across 38 files |
| `selector-max-type` | warning (`*.module.{css,scss}`) | nudge toward classes; 109 warnings, non-blocking |

## Notable decisions

- **`react-hooks/rules-of-hooks` was deliberately *not* enabled.** It produced
  230 errors, but ~201 are false positives inherent to Flow's architecture:
  custom hooks implemented as class static methods (`ActionModel.useNew`,
  `OverlayController.useNew`, …) and hook calls inside functions named `render`.
  The rule cannot model those and there is no config to whitelist them. The
  useEffect-dependency review comments that motivated it are actually covered by
  `exhaustive-deps` (enabled as warn).
  - The ~29 *genuine* findings it surfaced (early-return / conditional-throw
    before hooks) **are fixed** in `fix: call React hooks unconditionally…`
    (`RemoteRoot`, `RemoteRendererBrowser`, `useWatchPathname`,
    `AnchorNavigation`). Behaviour is preserved; hook order is made stable.
  - Two remaining findings are hooks called inside render-prop callbacks in
    **non-shipping** code — a demo (`apps/remote-dom-demo/.../remote/list/page.tsx`)
    and a test (`Form.browser.test.tsx`). They are functionally stable (the hook
    runs within the parent's render) and left as-is.

- **Logical properties scoped to inline direction only.** The full rule flags
  476 declarations, but ~376 are `width→inline-size` / `height→block-size` /
  block-axis sizing — aggressive and unrelated to the original review nudges
  (which were about left/right). The config excludes those; only left/right,
  `margin/padding/border-inline`, corner radii and `text-align: left/right` are
  enforced. LTR rendering is unchanged.

- **`selector-max-type` as a warning**, not an error: `selector-max-type: 0` is
  blunt for this codebase (React-Aria internals, `svg`, `input` often carry no
  class) and has no autofix, so it nudges without failing CI.

## Verification

- `pnpm lint` is green (0 errors; warnings only).
- `pnpm affected:test` (unit + compile vs `main`) passes.
- The logical-property changes are LTR-neutral, but a run of the visual suite
  (`run-visual-tests` label) is worth a sanity check since 38 stylesheets changed.


packages/components/src/components/List/components/ListItemView/ListItemView.module.scss
```diff
@@ -204,15 +204,14 @@
       :global(.flow--heading--heading-content) {
         position: absolute;
         top: var(--list-item--padding);
-        left: var(--list-item--padding);
-        right: var(--list-item--padding);
+        inset-inline: var(--list-item--padding) var(--list-item--padding);
```
Lisa18289: da müsste die var nicht doppelt rein, macht aber auch nix kaputt

mfal: Stimmt — auf `inset-inline: var(--list-item--padding)` reduziert (ein Wert gilt für beide Seiten). Kam durch den logical-property-Autofix rein.

packages/components/src/components/Switch/Switch.module.scss
```diff
@@ -41,6 +41,11 @@
     transform: translateX(0);
     position: absolute;
     top: var(--switch--track-padding);
+
+    /* Must stay physical: `.label-leading` sets `direction: rtl` as a layout
```
Lisa18289, line 45: oder sollte hier lieber der aufbau mit dem direction: rtl angepasst werden? das sollte ja eher für leserichtung als für css hacks verwendet werden

mfal: Guter Punkt — den `direction: rtl`-Hack entfernt: die Reihenfolge macht jetzt `flex-direction: row-reverse`, und der Handle nutzt wieder das logische `inset-inline-start` (ohne stylelint-disable). Layout unverändert; der Label-Text rendert jetzt korrekt LTR statt RTL-beeinflusst. Baselines aktualisiert.

Lisa18289 (approved):

packages/components/src/components/Switch/Switch.module.scss
mfal: Korrektur: Ich habe den row-reverse-Umbau wieder zurückgenommen. Das Entfernen von `direction: rtl` ändert das Rendering des Label-Texts (RTL→LTR) — also eine optische Änderung. Dieser PR soll rein Lint-/Logical-Property-Enforcement sein und **visuell neutral** bleiben, daher gehört der `direction: rtl`-Cleanup in einen eigenen PR. Hier bleibt es beim minimalen, deckungsgleichen Fix (physical `left` + stylelint-disable). Ich lege dafür einen Folge-Task an.

Lisa18289 (approved):

PR #2719, opened 2026-07-27
docs: components overview page 

Lisa18289:


mfal:
I like the overview and the component wireframes very much. Maybe the components should be sectioned into groups like in the menu as well.

maaaathis (approved):

Issue #2711, opened 2026-07-23
Release strategy for 1.0.0 — fast fixes, curated feature releases (accepted)


mfal:
# Requirement (binding)

Move Flow out of the `0.2.0-alpha.*` line to a stable **`1.0.0`**, and adopt a
release model where:

- **fixes ship as fast as possible**, and
- **features are collected and released together, going live in a bundle with a
  good, curated changelog.**

### Success criteria

- A fix can reach npm without waiting for any unreleased feature.
- A feature never ships on its own by accident; features accumulate and are
  released deliberately.
- Every feature release has a single, curated, human-readable changelog.
- `1.0.0` comes with a clear, documented statement of what is (and isn't) covered
  by the semver stability guarantee.

> **Status: accepted.** This model was proposed as an RFC and has been adopted
> by the team. The *requirement* above and the solution below are the agreed
> approach. Sub-topics are being captured as their own ADRs as they mature — the
> per-component lifecycle status is already fixed in **ADR 0003 (#2722)**.
> Items still genuinely undecided are collected under *Open questions* at the end.

---

# The model

## Glossary (ubiquitous language)

- **Stable line** — the `main` branch. Its HEAD is published to npm dist-tag
  `latest`. Only fixes land here directly.
- **Collection branch** — the `next` branch. Equals the Stable line plus all
  accumulated, not-yet-released features. Feature work targets it.
- **Major line** — an on-demand branch (e.g. `2.0`) that accumulates breaking
  changes toward the next Major. Only exists while a Major is maturing.
- **Forward-merge** — the automatic merge of a lower line into the line above it
  (`main → next → major line`), keeping higher lines a superset of lower ones
  without cherry-picking.
- **Promotion** — merging a higher line down into `main` to release it
  (`next → main` cuts a Minor; major line → `main` cuts a Major).
- **Release-PR** — a `next → main` (or major line → `main`) PR carrying the
  auto-generated, then hand-curated changelog. Its merge *is* the release moment.
- **Public API** — the surface under the semver guarantee (see the semver
  contract below).

## The core constraint

With Conventional Commits on a **single** branch, the moment one `feat:` commit
lands, the next release is forced to be a Minor — you can no longer ship "just the
fix" while a feature waits. "Fixes fast" and "features collected" are mutually
exclusive on one branch. This drives the whole design below.

## Proposed model — two standing lines + on-demand major line ("2 + n")

```mermaid
flowchart LR
    fix["fix:"] --> main
[...]

PR #2729, opened 2026-07-28
feat(components): add component lifecycle status registry (ADR 0003)

mfal:
## What

Implements the **foundation** of [ADR 0003 – Component lifecycle status](docs/adr/0003-component-lifecycle-status.md): a machine-readable, generated, committed **status registry** that derives a per-component lifecycle status (`{ level: "beta" | "stable" | "deprecated", isNew: boolean }`) from component-level JSDoc, exported via the `/internal` package entry.

The registry is the single source that will later drive documentation badges, Storybook badges, and the breaking-change guard (RFC #2711). With no component annotated yet, every component resolves to `stable` — zero migration cost.

## How

- **Derivation** (`dev/status-registry/deriveComponentStatus.ts`): component-level `@deprecated` → `deprecated` (wins over `beta`); `@flowStatus` tokens `beta`/`new`; default `stable`. The rule lives in exactly one place.
- **Keyed by public import specifier.** Each entry is `"<import-specifier>#<exportName>"` — e.g. `@mittwald/flow-react-components#Button`, `@mittwald/flow-react-components/nextjs#Link`. The registry covers **all five component-bearing export entries** (`.`, `./flr-universal`, `./nextjs`, `./react-hook-form`, `./password-tools`), not just the main surface — so integration components (and components public only via an integration, e.g. `PasswordCreationField` via `./password-tools`) are covered too.
- **Generator** (`dev/status-registry/generateStatusRegistry.ts`): reads `doc-properties.json` + `package.json` + the five entries' index files; per-entry regex parsers collect the exported component names; each `(entry, name)` resolves to its source component by matching `filePath` against the entry's source root (deterministic, absolute/relative-agnostic); status is derived and keyed by specifier. Writes two committed artifacts under `src/status/`:
  - `component-status.json` — canonical flat map, `jq`-readable (what the git-ref-comparing guard needs). Deprecated entries also carry a `deprecationNotice` (the migration text from `@deprecated`), so the docs Deprecated-callout reads it straight from the registry.
  - `componentStatus.ts` — typed wrapper: `flowComponentStatus` + `getFlowComponentStatus(name, entry?)` (defaults to the `.` surface).
- **nx wiring**: `build:status-registry` (`dependsOn: build:docs-properties`), wired into `build`; outputs guarded by the existing "Check all generated code is committed" CI step. Generated files excluded from `components-src` to avoid a cache self-invalidation loop.
- **Export**: registry + `getFlowComponentStatus` + types (`FlowComponentStatus`, `FlowExportEntry`) via `src/index/internal.ts`.

## Key decisions

- **Specifier keys, full contract.** A component is a contract unit *at a given export specifier*; `@mittwald/flow-react-components#Link` and `@mittwald/flow-react-components/nextjs#Link` are different units. Keying by specifier resolves same-name collisions across surfaces **structurally** (the earlier `displayName`-keyed design needed a fragile `src/integrations/**` path exclusion that broke on react-docgen's inconsistent absolute/relative `filePath`s — now deleted). `flr-universal` gets its own keys (status mirrors its `.` counterpart; consistent by construction). **Delivered: 133 keys** across the five entries.
- **`RouterProvider` granular & correct.** `@mittwald/flow-react-components/nextjs#RouterProvider` is tracked; `@mittwald/flow-react-components#RouterProvider` is absent (a react-aria re-export with no Flow doc — like `IntlProvider`/`Render`), so the Next.js integration component no longer masquerades as the core one.
- **Icons intentionally out of scope, per-icon.** Icons carry no JSDoc; their lifecycle lives in `icons-base/icons.yaml`. The `Icon` wrapper represents them (`stable`). **Icons are never removed — only deprecated**, so a breaking-change guard has no per-icon removal to detect and per-icon entries add no value. (ADR 0003 §4.)
- **Hardening**: a generation-time `level`-union guard (the `dev/`↔`src/` status types are deliberately duplicated — separate tsconfig `rootDir`s forbid cross-import), a fail-loud generator error when `doc-properties.json` is missing, and deterministic, byte-reproducible generation.

## Out of scope (deferred, per the ADR)

- The **Storybook** decorator and **`apps/docs`** badges/callouts/grouping — separate consumer plans; they import `/internal` and call `getFlowComponentStatus(name, entry)`.
- The **breaking-change guard** and its `beta`/`deprecated` exemption — RFC #2711.
- **Curating** which components become `beta` — the 1.0.0 API review.

## Testing

- `pnpm nx test:unit components` — **149/149 green** (derivation, the three per-entry parsers, the specifier resolver incl. the Link collision-split / RouterProvider granular / flr-universal duplication / absolute-vs-relative path resolution, and derivation edge cases).
- `pnpm nx test:compile components` — clean; the generated wrapper types `SPECIFIER_BY_ENTRY` as `Record<FlowExportEntry, string>`, so the entry table and the `FlowExportEntry` union can't silently drift.
- Reproducibility verified: `pnpm nx build:status-registry components` regenerates byte-identically (`git diff --exit-code` clean) and is a full nx cache hit.


packages/components/dev/status-registry/generateStatusRegistry.ts
```diff
@@ -0,0 +1,105 @@
...
+  );
+
+  for (const [name, status] of Object.entries(unsortedRegistry)) {
+    if (
+      !ALLOWED_LEVELS.includes(status.level as (typeof ALLOWED_LEVELS)[number])
+    ) {
+      throw new Error(
+        `Unknown status level "${status.level}" for component "${name}". ` +
+          `If you added a level to DerivedComponentStatusLevel (dev/), add it to ` +
+          `FlowComponentStatusLevel (src/status/types.ts) too.`,
```
maaaathis: We shouldn't explain how to add a new status at this point

packages/components/dev/status-registry/deriveComponentStatus.ts
```diff
@@ -0,0 +1,36 @@
...
+  const isDeprecated = "deprecated" in tags;
+
+  const level: DerivedComponentStatusLevel = isDeprecated
+    ? "deprecated"
+    : tokens.includes("beta")
+      ? "beta"
+      : "stable";
+
+  return { level, isNew: tokens.includes("new") };
+};
```
maaaathis, line 51: Could we make this more consistent and avoid the nested ternary? We extract `isDeprecated` into a variable, but check `tokens.includes("beta")` inline. Deriving both flags first and then using a straightforward if/else if would make the status priority easier to follow.

packages/components/dev/status-registry/deriveComponentStatus.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    { level: "deprecated", isNew: false },
+  ],
+  [
+    "empty @deprecated tag still deprecated",
+    { deprecated: "" },
+    { level: "deprecated", isNew: false },
+  ],
+  [
+    "deprecated wins over beta",
+    { flowStatus: "beta", deprecated: "gone soon" },
```
maaaathis, line 40: If we prefer the default `deprecated` annotation over the custom `flowStatus: "deprecated"` then we should discontinue the custom `deprecated` flow status

maaaathis (changes requested):

packages/components/dev/status-registry/generateStatusRegistry.ts
mfal: Trimmed the error message to just report the unknown level and component — dropped the how-to-add-a-status explanation.

packages/components/dev/status-registry/deriveComponentStatus.ts
mfal: Refactored: derive `isDeprecated`/`isBeta`/`isNew` up front, then a plain `if / else if` for the level (deprecated > beta > stable), no nested ternary.

packages/components/dev/status-registry/deriveComponentStatus.test.ts
mfal: Good prompt — there actually is no custom `flowStatus: "deprecated"` path: deprecation is derived solely from the standard `@deprecated` tag (ADR §3), and `@flowStatus` recognises only `beta`/`new`. I added a test pinning that `flowStatus: "deprecated"` is ignored (→ stable). The `deprecated`-wins-over-`beta` case stays, since ADR §3 mandates that conflict rule for a component mis-annotated with both.

packages/components/dev/status-registry/exportEntries.ts
```diff
@@ -0,0 +1,45 @@
...
+ * is the path fragment its components' source files live under (used to
+ * disambiguate same-named components across entries).
+ */
+export interface StatusExportEntry {
+  key: string;
+  indexFile: string;
+  sourceRoot: string;
+}
+
+export const STATUS_EXPORT_ENTRIES: StatusExportEntry[] = [
```
mfal, line 15: Add a hint in the appropriate doc to add an entry when new integrations are supported.

packages/components/src/status/componentStatus.ts
```diff
@@ -0,0 +1,30 @@
...
+import type {
+  FlowComponentStatus,
+  FlowComponentStatusRegistry,
+  FlowExportEntry,
+} from "./types";
+
+export const flowComponentStatus =
+  componentStatus as FlowComponentStatusRegistry;
+
+const SPECIFIER_BY_ENTRY: Record<FlowExportEntry, string> = {
```
mfal: This duplicates data already defined in exportEntries.ts

packages/components/src/status/component-status.json
```diff
@@ -0,0 +1,534 @@
...
+    "level": "stable",
+    "isNew": false
+  },
+  "@mittwald/flow-react-components/nextjs#Link": {
+    "level": "deprecated",
+    "isNew": false
+  },
+  "@mittwald/flow-react-components/nextjs#LinkProvider": {
+    "level": "deprecated",
+    "isNew": false
```
mfal: `depreactionNotice` is missing

packages/components/dev/status-registry/exportEntries.ts
mfal: Added a note in `packages/components/AGENTS.md` (right after the Public API surfaces table): a new integration export entry must be registered in `STATUS_EXPORT_ENTRIES` (`dev/status-registry/exportEntries.ts`) and the `FlowExportEntry` union (`src/status/types.ts`) — with the reminder that the zero-components generation guard and `test:compile` (via the `Record<FlowExportEntry, string>` net) catch omissions.

packages/components/src/status/componentStatus.ts
mfal: Removed the `SPECIFIER_BY_ENTRY` table. Since a `FlowExportEntry` token is exactly the specifier suffix, the wrapper now derives it from a single `PACKAGE_NAME` constant + a rule (`entry === "." ? PACKAGE_NAME : `${PACKAGE_NAME}/${entry}``), mirroring `specifierOf` in exportEntries.ts — no duplicated data.

packages/components/src/status/component-status.json
mfal: Added a `deprecationNotice` field to deprecated entries, sourced from the component-level `@deprecated` text (present only when deprecated with non-empty text). `nextjs#Link` and `nextjs#LinkProvider` now carry it, so the docs Deprecated-callout reads the migration text straight from the registry. Added to both `FlowComponentStatus` and its `dev/` twin.

packages/components/project.json
```diff
@@ -50,6 +57,26 @@
...
+      "inputs": [
+        "{projectRoot}/dev/status-registry/**/*",
+        "!{projectRoot}/dev/status-registry/**/*.test.ts",
+        "{projectRoot}/src/status/types.ts",
+        "{projectRoot}/package.json",
+        "{projectRoot}/src/components/public.ts",
+        "{projectRoot}/src/index/flr-universal.ts",
+        "{projectRoot}/src/integrations/nextjs/index.ts",
+        "{projectRoot}/src/integrations/react-hook-form/index.ts",
+        "{projectRoot}/src/integrations/@mittwald/password-tools-js/index.ts",
```
mfal:
```suggestion
        "{projectRoot}/src/integrations/*/index.ts",
```

mfal: Good call to consolidate — applied, but as `src/integrations/**/index.ts` rather than `*/index.ts`: the single-segment `*/` misses `@mittwald/password-tools-js/index.ts` (nested under the `@mittwald` scope), which would silently drop password-tools from the cache inputs. `**/index.ts` catches all three entry indexes; it also matches the component-barrel `index.ts` files, but that is harmless — the generator only parses the entry indexes, output is deterministic, and adding/removing a component already invalidates via `doc-properties.json`.

maaaathis (approved):

Issue #2738, opened 2026-07-29
1.0.0 release — tracking issue (fast fixes, curated feature releases)


mfal:
Tracking issue for everything that must happen to move Flow off the
`0.2.0-alpha.*` line to a stable **1.0.0** and stand up the release model.

Umbrella (accepted): #2711 — *fast fixes on `main`, curated feature releases via
`next`*. Detailed sequence with per-step verification lives in the cut plan
(kept out of git; ask a maintainer for the working copy). This issue is the live
checklist.

**Legend:** ✅ done · 🟠 in review (PR open) · 🚧 draft · ⬜ todo · ⏳ waiting (external/gated)

**Labels:** everything belonging to the 1.0 effort keeps `1.0`; the timing split
is [`pre-1.0`](https://github.com/mittwald/flow/labels/pre-1.0) (blocks the cut)
vs. [`post-1.0`](https://github.com/mittwald/flow/labels/post-1.0) (trails it).
Open `feat:` PRs are `post-1.0` by default — after the cut the routing guard
sends them to `next`.

> **Hard constraint (product management, 2026-08-04): 1.0.0 ships with NO
> breaking changes.** A major version number is not a licence to break, and an
> existing codemod is explicitly **not** a justification for breaking consumers.
> 1.0.0 is **additive-only** — new API and deprecations (`@deprecated` +
> `useWarnDeprecation`) are fine; removing, renaming, or relocating any existing
> public export waits for **2.0.0**. This scopes the pre-cut API review below and
> defers the breaking part of the export-surface RFC (#2774).

## Decisions & ADRs (gate the public contract)

- [x] ✅ Release model accepted — #2711
- [x] ✅ ADR 0001 CSS cascade layers · `docs/adr/0001-*`
- [x] ✅ ADR 0002 Stylesheet generation · `docs/adr/0002-*`
- [x] ✅ ADR 0003 Component lifecycle status → **Accepted** (#2722, merged in PR #2741)
- [x] ✅ ADR 0004 Forward-merge `main`→`next` → **Accepted** (#2727, merged in PR #2741)
- [x] ✅ ADR 0005 Semver contract at 1.0.0 (runtime API guaranteed; types/visual/DOM/CSS/token names+values not) → **Accepted** (merged in PR #2742). Folds in the **Node engines floor + React peer policy** from #2728, so a separate Node ADR is **not** needed.

## Release infrastructure (build)

[...]

PR #2747, opened 2026-07-29
ci: add commit-guard workflow (conventional PR title + routing)

mfal:
Adds the two commit-level gates of the 1.0.0 release model (#2711), in one
`commit-guard.yml` workflow. Tracked in #2738.

## 1. Conventional PR title
Lints the **PR title** with `amannn/action-semantic-pull-request` (pinned by
SHA, `v6.1.1`). Rationale: this repo **squash-merges**, so the PR title becomes
the commit subject that lerna derives versioning + changelog from — the title is
the thing that must be a valid Conventional Commit. (This is why the PR-title
linter is used over a commit-level one like `wagoid/commitlint-github-action`,
whose per-commit checks a squash discards.)

## 2. Routing guard (feature → next, not main)
On PRs targeting `main`, rejects `feat:` / `feat!:` / `BREAKING CHANGE` — those
belong on `next` or a major line.

**Self-gating:** the guard is **dormant until the `next` branch exists** (i.e.
until the 1.0.0 cut). Pre-cut, every change — features included — legitimately
targets `main`, so it skips with a notice and does not block current work. It
auto-activates when `next` is created at the cut. Promotion PRs (`head == next`
or a major line → `main`) are exempt, per ADR 0004 §8.

## Notes
- `amannn` swapped in for the `wagoid` example named during design — it lints the
  PR title (correct for squash) rather than individual commits. Happy to switch
  to commit-level linting if preferred.
- Follow-up (not here): the **version-contract guard** from ADR 0005 (fail a PR
  that raises `engines.node` or narrows a peer range without a breaking marker).


.github/workflows/commit-guard.yml
```diff
@@ -0,0 +1,111 @@
...
+          fail=0
+
+          # A feature belongs on `next`, not directly on `main`.
+          if printf '%s' "$PR_TITLE" | grep -qiE '^feat(\([^)]+\))?!?:'; then
+            echo "::error::PR title is a 'feat' — features target 'next', not 'main'."
+            fail=1
+          fi
+
+          # Any breaking change (type! marker, or a BREAKING CHANGE footer)
+          # belongs on the major line, never directly on `main`.
```
mfal: And not on next also

mfal: Fixed in e877490 — routing is now base-aware: breaking changes (`feat!`/`BREAKING CHANGE`) are rejected on PRs to `next` too (they belong on the major line), not only on `main`. `feat:` stays allowed on `next`.

maaaathis (approved):

PR #2749, opened 2026-07-29
ci: add run-cross-version-tests PR label workflow

mfal:
## What

Adds a label-triggered workflow that runs the cross-version smoke tests on a pull request on demand, analogous to the existing `run-visual-tests` label.

## Why

Both cross-version harnesses (iframe + in-process) currently run **only** in the scheduled visual workflow — not on PRs, because they network-install old published versions. That means a backwards-compatibility regression a change introduces is only caught by the next scheduled run, *after* the PR has merged. Removing a component or changing its rendered structure is exactly the kind of change that produces an old-vs-current divergence and may require adjusting the version gates.

## Changes

- **New workflow** `.github/workflows/test-cross-version-label.yml` — triggered by adding the `run-cross-version-tests` label (or `workflow_dispatch`). Mirrors the scheduled `cross-version` job: builds host packages + installs old versions (`prepare`), runs the iframe harness (with retry for connection flakes), then the in-process harness (`always()`). Removes the triggering label (re-add to re-run) and posts/updates a single per-version PASS/FAIL summary comment on the PR.
- **`packages/remote-react-components/CONTRIBUTE.md`** — new "Running them on a pull request" section: use the label whenever a component is removed or its rendered structure changes, since the version gates (`test.skipIf(crossVersion({ below }))` / `scenarioVersionSupport.ts`) may need adjusting.
- **`AGENTS.md`** — a Common-Failures row pointing at the label for the "scheduled cross-version tests go red after merge" trap.

## Follow-up (manual)

The `run-cross-version-tests` label must be created in the repo's GitHub label settings before it can be applied to a PR (same as `run-visual-tests` / `update-screenshots`).


maaaathis (approved):
glhf

PR #2750, opened 2026-07-30
fix(remote-react-components): stabilize flaky cross-version iframe suite

mfal:
## What

Fixes the frequently-red scheduled **cross-version** iframe smoke-test job (`test-visual-scheduled.yml`).

## Why it was flaky

The job failed ~40% of runs with `Could not establish remote connection: Timeout reached`. Investigation of the CI logs showed this is **not** an HTML/backwards-compat diff — the in-process harness passed every scenario in the same runs. It's the real `@quilted/threads` iframe handshake wedging a whole version's run under CI load: each run a *random* older/heavier version timed out on all its scenarios at once (60s each), and only a **fresh server pair** recovered it. The existing 3× whole-suite shell retry never converged, because a different version flaked on each attempt.

Two root causes:

1. **Double `globalSetup` → 4 Vite servers instead of 2.** Vitest browser mode runs `globalSetup` once *per project*, and a single-instance browser config still yields two projects (root + the `webkit` instance), so `setup()` fired twice and started **four** dev servers where two suffice. On the 2-core runner the two redundant servers force-optimizing heavy dep graphs starved the handshake past its timeout.
2. **Retry granularity mismatch.** Retrying the *entire* suite couldn't converge against a per-version, per-server-instance wedge.

## What changed

- **`e2e/cross-version/setupGlobal.ts`** — memoize a single server pair across both project setups and ref-count teardown, so exactly **2** servers start (verified in CI: 8 `serving version` logs for 4 versions, down from 16).
- **`dev/cross-version/crossVersionRunner.ts` + `run.ts`** — retry each version with **fresh servers** (`attemptsPerVersion: 3`) — the only thing observed to clear the wedge. A real diff still fails deterministically on every attempt, so retrying never masks a genuine break.
- **`test-visual-scheduled.yml`** — drop the coarse 3× whole-suite shell loop (retry now lives per-version in the runner); this also stops tripling the runtime.
- **`e2e/cross-version/setupGlobal.test.ts`** — new regression test locking in the shared-pair + ref-counted-teardown behavior.

## Verification

Dispatched run on this branch: [`cross-version` job green](https://github.com/mittwald/flow/actions/runs/30456641571) — 2 servers per iteration, 4/4 versions pass, **0** connection timeouts, **0** retries needed (halving contention alone cleared it this run; the per-version retry remains as a backstop). `test:compile`, ESLint, Prettier and the new regression test all pass locally.

**Caveat:** a single green run isn't statistical proof the flake is eliminated, but the *cause* (4→2 servers) is deterministically confirmed, and the retry covers the residual risk.

### DIFF
--- a/.github/workflows/test-visual-scheduled.yml
+++ b/.github/workflows/test-visual-scheduled.yml
@@ -122,26 +122,20 @@ jobs:
         # Renders each installed old remote version through the current host over
         # the real iframe connection and compares its host HTML to an ephemeral
         # reference rendered from the current version. WebKit-only. The
-        # @quilted/threads handshake intermittently times out under CI ("Could not
-        # establish remote connection: Timeout reached") — a transient infra flake,
-        # not a diff. Retry the whole step (same idea as the visual job's Firefox
-        # hang retry): a real HTML divergence fails deterministically on every
-        # attempt, while a connection flake clears on a fresh run. Runs the package
-        # script directly (not `nx run`) so a retry doesn't re-trigger the
-        # non-cached network prepare above; the final attempt's `failures` output
-        # feeds the Slack alert.
-        run: |
-          for attempt in 1 2 3; do
-            echo "::group::Cross-version smoke tests (attempt $attempt/3)"
-            if corepack pnpm --filter @mittwald/flow-remote-react-components test:cross-version; then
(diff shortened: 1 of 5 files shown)


maaaathis (approved):
glhf

PR #2752, opened 2026-07-30
ci: add version-contract guard (engines.node + peer ranges)

mfal:
Implements the **Version-contract guard** todo from the 1.0.0 tracking issue (#2738) — the ADR 0005 follow-up that "would have caught #2728".

## What

A CI guard that fails a PR which **raises `engines.node`** or **narrows a peer-dependency range** on a publishable Flow package without a breaking marker, per [ADR 0005](docs/adr/0005-semver-contract.md) §2/§3. It complements the existing `routing` job: `routing` enforces *where* a marked-breaking change may land; this guard detects a `package.json` change that *is* breaking but was shipped unmarked.

## How

- `​.github/scripts/version-contract-lib.mjs` — a pure, dependency-free classifier (`node:test`-covered). Compares each manifest's `engines.node` and peer ranges between the PR base and head; widening a peer / lowering-or-removing a Node floor is OK, narrowing a peer / raising-or-adding a Node floor is breaking. Fail-closed: any range outside the supported grammar counts as breaking.
- `​.github/scripts/version-contract-guard.mjs` — thin git-IO shell: diffs every `package.json` (base SHA vs working tree), calls the lib, sets the exit code. Passes with a breaking marker (`type!:` / `BREAKING CHANGE:`), fails without.
- A `version-contract` job in `commit-guard.yml`, gated **exactly like `routing`**: it is **dormant until the `next` branch exists** (so it's a no-op on CI today, pre-1.0.0-cut) and exempts promotion/sync heads.
- An ESLint override so the new Node scripts lint clean.

## Scope decisions (v1)

- Watches **all peer dependencies** (caught #2728's `react-hook-form` `*` → `^7.65.0`), not just `react`.
- **Marker-only** escape hatch — no EOL-Node exception yet (documented extension point).
- **New peers are unflagged** in v1.
- Scope = `@mittwald/`-scoped, non-private packages (handles the `private: true` vs `"true"` inconsistency).

## Notes

- **No-op on this PR** — `next` doesn't exist yet, so the job self-gates to dormant. It activates at the cut.
- Post-cut, the self-gating dormancy in `routing` + this job becomes removable (tracked as a post-cut cleanup item in #2738).

Refs #2738.


maaaathis (approved):
glhf

PR #2757, opened 2026-07-30
feat(components): warn on all remaining deprecated APIs via useWarnDeprecation

mfal:
## What

Several deprecated APIs were marked `@deprecated` in JSDoc and still supported at runtime, but emitted **no** deprecation warning — so consumers had no signal they were on a legacy path. This wires `useWarnDeprecation` into every deprecated API that has a render/hook entry point:

| API | Kind | Where |
| --- | --- | --- |
| `Select.onSelectionChange` | prop | warns when set; JSDoc now points to `onChange` |
| `ItemViewShape.fallback` | List item prop | warned in the `List` render-time constructor |
| `ListLoaderAsyncResource` | component | warns on render → `ListLoaderHooks` |
| nextjs `Link` | component | warns on render → `RouterProvider` |
| nextjs `LinkProvider` | component | warns on render → `RouterProvider` |
| `generatePasswordCreationFieldValidation` | hook alias | wrapped to warn, then delegates to `usePasswordCreationFieldValidation` |

All messages follow the established shape: `The '<name>' <prop|component|function> is deprecated and will be removed in a future release. Use '<replacement>' instead.`

## Why

`useWarnDeprecation` is a general-purpose deprecation mechanism (plain React context + `console.warn` + optional `onWarning` callback) — **not** tied to flr-remote or `@flr-generate`. It had only ever been applied to a handful of extension-facing props, leaving equivalent non-remote/integration deprecations silent. This closes those gaps.

### Notable detail — `ItemViewShape.fallback`

The warning sits in the `List` model constructor, not the `ItemView` constructor. `List` runs during render (it already calls `useSettings()` and `Model.useNew` hooks), so a hook is valid there. `ItemView` is constructed **conditionally** (`itemView ? new ItemView(…) : undefined`), which would violate the rules of hooks — so we warn one level up where the shape enters render.

## Docs

The hook was documented exclusively under the `@flr-generate` / remote sections, which reads as if it only applies to extension-facing props. Clarified in the root & components `AGENTS.md` and `PATTERNS.md` that it is general-purpose, plus the render/hook-context constraint that dictated the `List` placement above.

## Not changed

Type-only deprecations (`CartesianGridProps`, `CartesianChartEmptyViewProps`) stay unwarned — no runtime path to warn on.

## Verification

- ✅ `pnpm nx test:compile components` (tsc)
- ✅ `eslint` on all touched files (0 errors)
- ✅ `prettier --check` (incl. markdown)
- ✅ `pnpm nx test:unit components` — 156 passed

No `@flr-generate` prop *types* changed, so no generated code needed regeneration.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,7 +52,9 @@ RemoteRoot + remote React components  ───►  RemoteRenderer + RemoteRecei
   developers.** Avoid breaking changes. When an API must change, keep the old
   path working and log usage with `useWarnDeprecation` (from
   `DeprecationWarningProvider`) so extension developers can be informed about
-  deprecation paths they still use.
+  deprecation paths they still use. `useWarnDeprecation` is the standard way to
+  flag any deprecated runtime path — see
+  [Deprecating an API](packages/components/AGENTS.md#deprecating-an-api).
 - Inside components that are part of the `flr-universal` export surface, compose
   other Flow components through their **views** (`@/views/*`) — views
   automatically switch to the remote counterpart in a remote context.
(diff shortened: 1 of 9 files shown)


maaaathis (approved):
lgtm

PR #2765, opened 2026-07-31
feat(remote-react-renderer): report component usage to the host

maaaathis:
## What & why

Report used components to the remote-react-renderer (mStudio flow-remote host)

## Base branch & title

This repo **squash-merges**, so your **PR title becomes the release commit** —
it must be a valid [Conventional Commit](https://www.conventionalcommits.org/)
(e.g. `fix(Button): correct focus ring`). A CI guard lints the title _and_ that
it matches the base branch. Pick the base by change type
([details](https://github.com/mittwald/flow/blob/main/CONTRIBUTE.md#choosing-the-base-branch)):

- `fix:` / `docs:` / `chore:` / `refactor:` / … → base **`main`**
- `feat:` (new feature) → base **`next`**
- Breaking change (`feat!:` / `BREAKING CHANGE:`) → base **the major line**

> Before the 1.0.0 cut the `next` / major lines don't exist yet, so everything
> targets `main` and the routing guard stays dormant.

## Checklist

- [ ] PR title is a Conventional Commit and matches the base branch above
- [ ] `pnpm lint` is clean and `pnpm affected:test` passes (browser tests if
      behavior changed)
- [ ] **Generated code is committed** (`git diff` is empty after the relevant
      `build:*` targets)
- [ ] User-facing strings added to **both** `de-DE` and `en-US` locale files
- [ ] Docs updated if a public API changed; intentional visual changes get
      updated snapshots / the `update-screenshots` label


mfal:
Two things — one about what this actually measures, one for a possible rework.

### This reports host output, not remote usage

Collecting in `createFlowRemoteComponentRenderer` means we count what falls out
on the host at the end — the `flr-*` elements that get materialized — not what
the extension author used. Those two diverge as soon as a self-contained remote
component renders component views of its own: the views are counted as well,
even though the contributor strictly speaking never used them directly. So
"used components" really means "components that appeared in the rendered output
tree", transitive composition included.

Whether that's a problem depends on the question the signal is supposed to
answer:

- _"Which Flow surface does this extension end up depending on / what breaks if
  we change X?"_ → host output is a defensible answer.
- _"Which components did the contributor actually use?"_ → that has to be
  collected in the remote, where the distinction between direct use and internal
  composition still exists.

Right now the naming (`onComponentUsage`, "Used components") reads like the
second, while the implementation delivers the first. Whichever way we go, this
belongs next to the existing "absence is not proof of non-usage" caveat in
`docs/remote-ui.md` — it's the same class of caveat and just as easy to trip
over.

### If it gets reworked to report from the remote

Please don't bolt another bespoke method onto `HostExports`. We already have
`setIsReady`, `setNavigationState`, `reportDeprecation`, … — every new
remote→host signal widens the versioned protocol contract and has to be
negotiated across versions forever.

Worth thinking about one generic event channel instead, roughly:

```ts
reportEvent: (event: { event: "ComponentRendered"; data: object }) =>
  Promise<void>;
```

with the payload per event type defined as a zod schema (validated on the host,
so an unknown or newer event coming from a remote can simply be ignored).
Adding a new event is then a schema, not a protocol change.

maaaathis:
@mfal 
Both points are in. Collection moved into the remote, and it now goes through a
generic event channel rather than a new `HostExports` method.

### You were right, and in both directions

I reproduced your case on our own demo page and then found the mirror image of
it. `apps/remote-dom-demo/.../remote/context-menu/page.tsx` writes
`Section, Header, Heading, ModalTrigger, Button, Modal, Content, ActionGroup,
Action, ContextMenuTrigger, ContextMenu, MenuItem, IconApp, Text`; the host-side
signal reported `Button, ClearPropsContext, Content, ContextMenu,
ContextMenuTrigger, DialogTrigger, Header, Heading, OverlayContent, Section`.

- **Over-count, exactly as you described:** `DialogTrigger`, `OverlayContent`,
  `ClearPropsContext` came only from inside `ModalTrigger`/`Modal`.
- **Under-count, which I had not considered:** `Modal`, `ModalTrigger`, `Action`,
  `ActionGroup` were missing entirely. There is no `flr-modal` — those
  `flr-universal` components execute *in the remote* and never produce an
  element. That covers all 16 host-only exports in `flr-universal.ts` (`Modal`,
  `List`, `LightBox`, `Popover`, `Action`, `NotificationProvider`, …).

So host output was not a coarser view of usage but a different set, and the
reading you were willing to accept ("what does this extension depend on") wasn't
served either: the list would have said "no extension uses `Modal`".

### What it looks like now

Collection sits in `flowComponent`, which is the single funnel for both the
generated remote components (`createFlowRemoteComponent` wraps them in it) and
the `flr-universal` components that run inside the iframe. `RemoteRoot` collects,
deduplicates per session, and reports.

On the demo's `/host/list` page: **13 components used by the extension** vs. **16
composed by Flow**. The old signal would have merged all of them into one list
and still omitted `List` and `Modal`.

### Direct vs. internal: the seam is the view, the mechanism is a prop

I first built this as a context flag set by the view and reset by the component.
That is wrong, and the tests showed it: between a view and the component it
resolves to there can be any number of non-Flow components, so the value kept
flowing down and flagged the consumer's own children as internal
(`Button:internal`, `Content:internal` for components the author wrote). It is a
**prop** now — props reach exactly one instance — and `flowComponent` strips it
before anything else sees it.

Second trap in the same area: `RemoteRoot` marks only view targets that are Flow
component names. `DialogTrigger` is not one, so it stays a bare remote element,
and remote-dom's slot handling passed the marker straight on to the author's
children.

Gaps that remain, all documented in `docs/remote-ui.md` next to the "absence is
not proof of non-usage" note:

- Components that are not `flowComponent`s report nothing — `ModalTrigger` and
  the other plain wrappers around `OverlayTrigger`.
- Icons surface as an internally composed `Icon` (ADR 0003 §4 does not track them
  per icon).
- Internal composition is recognized at the view seam, which is complete only as
  long as `flr-universal` components compose through views — the rule anyway.

### The channel

`reportEvent` with a zod schema per event type, validated on the host; an event
it cannot parse (unknown type, or a newer payload) is dropped, never thrown back.
Two things I wrote into the doc comment rather than leaving implicit:

1. `reportEvent` is itself a member older hosts lack, so callers use
   `reportEvent?.(…)` and swallow the rejection, like `reportDeprecation`.
2. Versioning moves rather than disappears: unknown event *types* are free,
   a changed payload for a known type is the same problem one level down. Extend
   schemas additively; a changed field means a new event type.

`zod@^4.4.3` is a new runtime dependency in `remote-core` (already present in
`components` and `ext-bridge`).

### Tests

- e2e over a real iframe and a real connection: `Modal` is reported (the
  under-count), `Option` from `CountryOptions` is internal while `Select` is
  direct, status resolution, once per session.
- Semantics unit-tested in `packages/components`, including that the children of
  an internally composed component are attributed to the consumer.
- Schema tests in `remote-core` for the unknown-event and invalid-payload paths.

mfal:
I think there is a conceptual error in how "internal composition" is determined
here. The special handling (ignore it) hinges on `isFlowComponentName()`, and
that predicate is the wrong tool in two independent ways.

## 1. `isFlowComponentName()` does not mean "internal composition"

It only says that the name is registered in `propsContextSupportingComponents` —
i.e. that the component is produced by the `flowComponent()` factory. Nothing
about that is related to composition.

What makes this concrete: `createFlowRemoteComponent` uses the *very same*
predicate to decide whether a remote component gets wrapped in `flowComponent`
at all:

```ts
if (isFlowComponentName(flowComponentTag)) {
  return flowComponent(flowComponentTag, …, { isRemoteComponent: true });
}
return element;
```

So the filter in `RemoteRoot`

```ts
isFlowComponentName(name) ? markInternalComposition(component as never) : component
```

is really answering _"will this map entry strip the marker prop again?"_ —
because `markInternalComposition` on anything else would push
`__flowInternalComposition` onto a raw `flr-*` element that has no
`takeInternalCompositionMark` to consume it. That is a prop-leak guard wearing
the costume of a classification, and the criterion it uses has nothing to do
with composition.

## 2. There are no "internal compositions" as an entity

There is no such thing as an internally-composed _component_ — there are only
_renders_ that originate from Flow's own composition, and the mechanism for that
is the **views**. So the rule should simply be: **a render that arrived through a
view is not tracked**, because it genuinely comes from a composition. `docs/remote-ui.md`
in this PR already says as much ("Internal composition is recognized at the
**view** seam") — the implementation just doesn't put the decision there. It
puts it in a rewritten view-component map plus a marker prop threaded through
`flowComponent`'s props, and then invents a category on top.

If the rule is "views aren't tracked", the flag disappears from the public
surface too: no `isInternalComposition` on `ComponentUsageEvent`, no
`markInternalComposition` / `takeInternalCompositionMark` /
`internalCompositionProp` in `internal.ts`, no "Composed by Flow" column — those
renders are simply not reported. And it would then behave identically in a local
and a remote render, which the current mechanism does not: locally the views fall
back to the real component (`useContext(viewComponentContext)["Button"] ?? Button`),
nothing is marked, and internal composition silently counts as consumer usage.

The very first browser test in this PR shows the hole:

```tsx
<Section>
  <Heading>Death Star</Heading>
  <Button>Fire</Button>
</Section>
// → ["Button", "Heading", "Section", "Text"]
```

That `Text` is nobody's consumer usage — it is `Button` composing
`<Text className={styles.text}>{children}</Text>` directly (`Button.tsx:190`),
past `TextView`. Not marked, so it lands in the consumer bucket. The PR docs do
concede this ("a component reaching past its view would look like consumer
usage"), but the test asserts the wrong outcome as if it were correct.

## Fallout worth knowing before the rework

- **54 of the 131 generated remote components are not `flowComponent`s** and
  therefore report nothing at all, marked or unmarked: `Table`, `TableBody`,
  `TableCell`, `TableRow`, `TableColumn`, `Skeleton`, `SkeletonText`, `Truncate`,
  `Flex`, `Div`, `Separator`, `LoadingSpinner`, `Markdown`, `Tooltip`,
  `OverlayContent`, `PopoverContent`, `Breadcrumb`, `BigNumber`, the whole chart
  family, … `docs/remote-ui.md` sells this as "`ModalTrigger` and the other plain
  wrappers around `OverlayTrigger`", which undersells it considerably — it is 40%
  of the remote surface. Side note: `componentUsage.test.ts` builds its fixtures
  out of exactly those (`OverlayContent`, `TableCell`), i.e. events the system
  cannot emit.
- The `as never` and `as unknown as FlowViewComponents` casts around the map
  rewrite are what let the mismatch pass unnoticed. Whatever replaces it should
  not need a cast.
- **Constraint for the rework:** whatever carries the "this came from a
  composition" information has to reset for children. Extension JSX passed as
  children into a Flow-composed component must still count as consumer usage —
  the existing test _"attributes the children of an internally composed component
  to the consumer"_ pins that, and a naive "we are inside a view" context would
  break it, since React context applies to the whole subtree.

mfal (changes requested):
See latest comment

mfal:
Rework looks right — view seam, context instead of prop, bare-element branch.

One gap: `"reports the components that render"` moved from exact `toEqual([...])`
to `arrayContaining([...])`, which drops `Text`. But `Text` is still reported —
`Button` renders `<Text>` directly instead of `TextView` (`Button.tsx:190`). The
one knowingly incomplete spot of the view-seam rule is now only in the docs, in
no test. I'd pin it:

```tsx
test("over-reports a composition that bypasses its view", async () => {
  // Known limitation, pinned on purpose: the exclusion sits at the view seam,
  // so Button rendering <Text> directly instead of TextView lands in the
  // consumer's bucket. Polling on Button is the flush gate — React runs the
  // child's mount effect first, so once Button is in, Text is too.
  const events = await renderCollecting(<Button>Fire</Button>);

  await expect.poll(() => componentsOf(events)).toContain("Button");
  expect(componentsOf(events)).toContain("Text");
});
```

Green on `6de16984` in WebKit and Firefox. Positive assertion on purpose —
`not.toContain` would be the wrong way round and would pass vacuously if
reporting broke.

mfal (approved):

PR #2766, opened 2026-08-03
fix(CodeEditor): render label above the editor and align padding

Lisa18289:
## What & why

In the `CodeEditor`, the `Label` was rendered **inside** the editor box, and the
editor had no inline padding at all, so line numbers and code stuck to the
border instead of matching the padding of every other form field.

**Label & field description** were passed through as `children` of `CodeMirror`.
They are now tunneled out of the editor — the same mechanism `useFieldError`
already uses for the field error — so the DOM order is `label → editor →
description/error`. Deliberately *not* solved by moving all `children` out:
`CodeBlock` passes its "show more" button as `children` and needs it rendered
inside the bordered box, on top of the fade gradient. With the tunnel,
`CodeBlock` is untouched.

**Padding** now comes from the `formControl` mixin. Because the highlight of the
active line has to span the full width, the inline padding is applied by
`.cm-line` and the elements of the first gutter instead of the editor box:

- code and line numbers sit at `padding-x` + border (`17px`), block at `8px` —
  measured identical to `TextArea`
- `.cm-activeLineGutter` starts at `1px` and `.cm-activeLine` ends at `1px`, so
  the band runs from border to border
- the now-pointless `border-radius` rules on `.cm-gutters` are gone: the band
  starts exactly `8px` from the edge, where the `8px` corner radius ends

Two related defects fixed along the way:

- **`isInvalid` had no visual effect.** The mixin styles the invalid state via
  `[data-invalid]`, but only `aria-invalid` was set — on the wrapper div.
- **Label and field description were not associated with the editor.** The
  editable element is `.cm-content` (`role="textbox"`), not the root element the
  props land on, so `aria-labelledby` / `aria-describedby` / `aria-required` /
  `aria-invalid` are now applied through the content attributes facet. Both are
  rendered into slot elements that carry the id, so the references never dangle
  when no label exists.

## Verification

- new `CodeEditor.browser.test.tsx`: label outside the editor + accessible name,
  accessible description, invalid state (`aria-invalid` + `data-invalid`)
- visual test extended by a labeled and an invalid instance; **darwin**
  baselines (webkit + firefox) for `CodeEditor` and `CodeBlock` regenerated
- `test:compile`, `test:unit components` (156), `test:browser components` (211),
  `lint` / `format` clean, generators produce no diff
- `CodeBlock` (default, truncated, children) checked in Storybook and via
  screenshots — unchanged apart from the code now getting the same `16px`
  padding its `--code-block--padding` token already used for the children path

`update-screenshots` label is set so CI regenerates the `-webkit-linux`
baselines.


packages/components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -74,20 +76,59 @@ export const CodeEditor = flowComponent("CodeEditor", (props) => {
...
 
   useMakeFocusable(localRef);
 
   return (
     <div className={rootClassName}>
-      <PropsContextProvider props={fieldPropsContext}>
+      <PropsContextProvider props={propsContext}>
+        <div className={styles.labelSlot} id={labelId}>
+          <UiComponentTunnelExit id="label" component="CodeEditor" />
+        </div>
```
mfal: You *can* use the render function of the TunnelExit to render the wrapping div coditionally.

packages/components/src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -124,6 +164,9 @@ export const CodeEditor = flowComponent("CodeEditor", (props) => {
             {children}
           </CodeMirror>
         </FieldErrorCaptureContext>
+        <div className={styles.fieldDescriptionSlot} id={descriptionId}>
+          <UiComponentTunnelExit id="fieldDescription" component="CodeEditor" />
+        </div>
```
mfal: See comment above.

packages/remote-react-components/src/tests/visual/__screenshots__/CodeBlock.browser.test.tsx/CodeBlock-firefox-darwin.png
mfal, line 1: The padding on the left doesn't look right. Is it?

mfal (changes requested):

packages/remote-react-components/src/tests/visual/__screenshots__/CodeBlock.browser.test.tsx/CodeBlock-firefox-darwin.png
Lisa18289: It is intentional — the left inset of the field is the same as on every other form control.

Measured (webkit, darwin):

| | inset from the border |
|---|---|
| `TextArea` text | 17px (`padding-inline` 16 + border 1) |
| `CodeEditor` / `CodeBlock` first column | 17px |

So the content box starts exactly where a `TextArea`'s text starts. What sits in that first column here is the line-number gutter, and the code follows after it. `--code-block--padding` is `size-px.m` as well, so the two `CodeBlock` variants in this screenshot now line up at the left edge — before this PR the code path had **no** inline padding at all (the line numbers were glued to the border, which is what this PR fixes).

The one number that is new beyond the field padding is the gap between the gutter and the code: it is `--form-control--padding-x` (16px), because the inline padding sits on `.cm-line` so that the highlight of the active line can still span the full width of the editor. In the screenshot that adds up to `1 + 16 + gutter + 16`.

If you'd rather have the code closer to the line numbers, the knob is that one declaration — gated so the gutterless case keeps the full field padding:

```scss
& :global(.cm-line) {
  padding-inline: var(--form-control--padding-x);
}

/* with a gutter, the field padding is already applied by the gutter */
&:has(:global(.cm-gutters)) :global(.cm-line) {
  padding-inline-start: var(--form-control--spacing-x); /* 8px */
}
```

Happy to push that if you prefer the tighter look — otherwise I'd leave it as is.

packages/components/src/components/CodeEditor/CodeEditor.tsx
Lisa18289: Good point — the wrapper is gone now, and without needing the render function at all.

The div only ever existed to carry the id that `aria-labelledby` / `aria-describedby` on the editor content point to (CodeMirror's editable element is `.cm-content`, so the attributes go through `EditorView.contentAttributes`, and they need something stable to reference). Putting those ids on the `Label` and the `FieldDescription` themselves — via the props context that already configures them — does the same job with **no** extra element:

```tsx
Label: { ...fieldPropsContext.Label, id: labelId, tunnel: { … } },
FieldDescription: { ...fieldPropsContext.FieldDescription, id: descriptionId, tunnel: { … } },
```

The field root now renders exactly `label` → editor → description, so the label and the description are direct flex items just like in every other form control. That also fixed a 1–2px offset I hadn't noticed: inside the wrapper the description was laid out as an inline box in a line box, as a flex item it gets blockified — the visual baseline is updated accordingly.

One consequence worth naming: when no label (or no description) is passed — `CodeBlock` is the common case — the reference has no target and the editor stays unnamed, exactly as before this PR. The render-function variant would behave the same way; the alternative would be to keep an always-present anchor element, which is what I had. Happy to go back if you'd rather not have a reference that can point at nothing.

packages/components/src/components/CodeEditor/CodeEditor.tsx
Lisa18289: Same as above — the wrapping div is gone, the id sits on the `FieldDescription` itself now (`673899d5e`).

mfal (approved):

PR #2800, opened 2026-08-05
test(remote-react-components): stop parking the pointer on the content

mfal:
## The bug

`setNeutralPointerPosition()` in `src/tests/lib/environments.tsx` called `rootContainerLocator.unhover()`. In the vitest browser locator, `unhover` is implemented as:

```js
unhover(e) { return this.triggerCommand('__vitest_hover', 'html > body', e) }
```

It *hovers the body*, and hovering targets an element's centre — so it parked the pointer at the viewport centre, which is where scenarios put their content. The helper produced the exact hover state it is named after preventing.

Thanks @mfal for spotting it on the `IllustratedMessage` baseline.

## Measured, not guessed

Sampling the soft secondary button's fill in `IllustratedMessage content` (resting is `--color--gray--400` = `#DCE9F7`, hover is `--color--gray--500` = `#B7C9DB`):

| baseline | fill | state |
| --- | --- | --- |
| `-webkit-darwin` on `main` | `#DCE9F7` | resting |
| `-webkit-linux` on `main` | `#B7C9DB` | **hover** |
| after this fix | `#DCE9F7` | resting |

The committed Linux baselines carry the hover. The darwin ones predate the leak — and a plain baseline refresh would have propagated it to them, which is what the first revision of this PR did before the cause was found.

## The fix

Hover a coordinate inside the container's 32px padding, which no scenario paints into. `force` skips the hit-target check: with a full-screen overlay open there is no free point in the viewport at all, and the pointer landing on a backdrop is harmless — all that matters is that it is off the components.

Rejected alternatives: an invisible element to hover fails under `Modal` and `LightBox`, which portal into `document.body` and intercept the hover from a different stacking context; there is no API to move the pointer to a bare coordinate.

## Baselines

- **17 `-webkit-darwin` baselines** refreshed, because they had drifted from what a current macOS/WebKit renders — text and icon rendering in components unrelated to any code change.
- **`IllustratedMessage-content-webkit-darwin` needed no refresh at all** once the pointer stopped hovering the button. It is not in this diff.
- **2 new** darwin baselines: `ActionGroup-preserveOrder` and `ActionGroup-size` had none, only the Linux ones.
- **`-webkit-linux` baselines are regenerated on CI** via the `update-screenshots` label. Expect the hover to disappear from every scenario where the viewport centre happened to sit on an interactive element — worth a look, since that tells you where a baseline was asserting a hover state nobody intended.

## Caveat

Darwin baselines encode one specific macOS/WebKit version and will drift again for a colleague on another. That is inherent to committing per-platform screenshots and is not addressed here.


packages/remote-react-components/src/tests/visual/__screenshots__/IllustratedMessage.browser.test.tsx/IllustratedMessage-content-webkit-darwin.png
mfal, line 1: The Secondary button looks like it is beeing hovered. But how could it be hovered? Maybe the virtual pointer is above the button when the screenshot is taken? If so, pointers should be moved to neutral position first.

mfal: Good catch, and the cause is worse than a stray pointer — but the hover is in the **linux** baseline, not this darwin one.

Resolving the tokens for `variant="soft" color="secondary"` in the light theme: resting is `--color--gray--400` = `#DCE9F7`, hover is `--color--gray--500` = `#B7C9DB`. This darwin file shows the light `#DCE9F7` (resting, correct); the committed `IllustratedMessage-content-webkit-linux.png` shows `#B7C9DB` — hover. The previous darwin baseline looks identical to this one, so regenerating changed nothing about the button state.

Your suspected mechanism is right, and it is `setNeutralPointerPosition()` itself. In `src/tests/lib/environments.tsx` it calls `rootContainerLocator.unhover()`, and in the vitest browser locator `unhover` is implemented as:

```js
unhover(e) { return this.triggerCommand('__vitest_hover', 'html > body', e) }
```

It *hovers the body*, and Playwright targets an element's centre — so it parks the pointer at the viewport centre (640, 360). The Secondary button occupies roughly x 530–648, y 332–368, which contains that point. The helper actively creates the hover it is named after preventing.

Not fixing it in this PR: the fix belongs in the harness, and it will change every linux baseline where the viewport centre lands on an interactive element — a wider change than a baseline refresh. Filing it separately unless you want it here.

mfal: Correction to my previous reply: I claimed the hover was only in the linux baseline and that refreshing changed nothing about the button state. Measuring the pixels instead of eyeballing them shows otherwise — sampling the button fill at (545, 350):

| baseline | fill | state |
| --- | --- | --- |
| `-webkit-darwin` on `main` | `#DCE9F7` | resting |
| `-webkit-darwin` as I had refreshed it here | `#B7C9DB` | **hover** |
| `-webkit-linux` on `main` | `#B7C9DB` | **hover** |
| after the fix now pushed | `#DCE9F7` | resting |

So the refresh *did* bake the hover into the darwin baseline. The darwin file on `main` simply predates the leak; the two shades are close enough that they look identical side by side, which is how I got it wrong.

The cause is as described: `unhover()` is "hover `html > body`", and hovering targets an element's centre. Fixed by hovering a coordinate inside the container's padding with `force`. `IllustratedMessage-content-webkit-darwin` is no longer in this PR's diff at all — it only ever differed because of the hover.

The `update-screenshots` label is set, so the linux baselines get regenerated with the pointer parked properly.

Lisa18289 (approved):

Lisa18289 (approved):

PR #2811, opened 2026-08-06
feat(Tooltip): show tooltips after 400ms

### LINKED ISSUE #2803: The display time of the tooltip in the Tooltip component feels unusually long
Image

Lisa18289:
## What & why

Closes #2803 — the tooltip delay felt unusually long. React Aria defaults to
1500ms; Flow already advises against tooltips on self-explanatory elements, so
the tooltips that remain carry information the user is waiting for.

Instead of just lowering the number, `delay` is now a named variant. Free-form
millisecond values invite one-off timings, and two cases cover what we actually
have:

- `"default"` (400ms) — elements that cannot be understood without the tooltip,
  e.g. icon only buttons
- `"long"` (1500ms) — supplementary information on elements that are already
  labeled

`closeDelay` stays a plain React Aria passthrough.

**Type-level break:** `delay={800}` no longer typechecks. It never was a
hand-written prop — it came in through the React Aria props spread — and
nothing breaks at runtime: an unknown value resolves to `undefined`, so Aria
falls back to its own default. The tooltip still works, it is just slow again.

### Docs

- New `# Delay` section in the tooltip overview with a live example, in the
  style of the Button variants page.
- Fixed a best practice that contradicted `"long"`: it forbade tooltips on
  labeled elements outright, while the "Verwendung" section right below asks
  for exactly that. The intended rule is that a tooltip must not *repeat* the
  visible label.

## Testing

`test:compile`, `test:unit` and `pnpm lint` are clean; the generators produce
no diff. Visual tests could not run locally (Playwright never connects to a
browser session in my environment) — added the `run-visual-tests` label, since
`Tooltip.browser.test.tsx` lost its 1500ms-era fixed wait.

### DIFF
--- a/apps/docs/src/content/04-components/overlays/tooltip/examples/delay.tsx
+++ b/apps/docs/src/content/04-components/overlays/tooltip/examples/delay.tsx
@@ -1,0 +1,19 @@
+import {
+  Button,
+  IconEdit,
+  Tooltip,
+  TooltipTrigger,
+} from "@mittwald/flow-react-components";
+
+<>
+  <TooltipTrigger>
+    <Button aria-label="Bearbeiten" variant="plain">
+      <IconEdit />
+    </Button>
+    <Tooltip>Bearbeiten</Tooltip>
+  </TooltipTrigger>
+  <TooltipTrigger delay="long">
+    <Button variant="plain">Bearbeiten</Button>
+    <Tooltip>Name und Beschreibung ändern</Tooltip>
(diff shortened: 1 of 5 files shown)


Lisa18289:
The chart itself needs Isolation:isolate, to stop the tooltip from overlapping other elements:

Bildschirmfoto 2026-08-06 um 13 58 34

mfal (approved):

PR #2817, opened 2026-08-06
chore(dev): fixes multiple development issues

ins0:
this fixes

- [x] generate `.scss` d.ts files for type safety
- [x] adds `/lib/scss/selectors.ts` to make dynamic classNames type safe
- [x] add scss.d.ts genration to `nx` and `vite` to keep `scss` and `scss.d.ts` generation always in sync 
- [x] fixes storybook freezes duo to huge file reads/writes - ts checks in storybook will use Typescript7 now
- [x] add a slim `tsconfig.ts` for storybook to resolve storybook ts errors/warnings


mfal (changes requested):
The freshly generated CSS-Module TS-stubs should be Prettier-ignored.

mfal:
### `dev/scss-types/generateScssTypes.ts` — two gaps that can flip CI's `git diff --exit-code`

Root cause: `vite-plugin-sass-dts` writes each `.d.ts` in a **fire-and-forget** callback that `transform` never awaits. No completion signal → `waitForWrites` (L57–92) falls back to an mtime heuristic with two holes.

**1. Race: `waitForWrites` returns before a slow write lands**
- All ~118 transforms fire back-to-back (L116–118), then we wait.
- Done condition (L80): all files exist **AND** mtimes stable ≥ `stableMs` (1500ms).
- If one Sass compile lands >1500ms after all others settled → loop returns success, `run()` resolves, process exits, pending `fs.writeFile` may never flush.
- → stale/missing declaration, non-deterministic under load.

**2. Silent stale declaration when an existing module stops compiling**
- Doc comment (L51–55) claims a failed compile fails the build — only true for a **new** module (`.d.ts` never appears → `allExist` false → 180s timeout → throw).
- Existing module with a committed `.d.scss.ts` that now breaks: sass-dts swallows the error, file never rewritten, `allExist=true`, mtimes stable → `waitForWrites` returns **success**, keeps the stale file. No error surfaces.

**Fix direction**
- Replace the mtime heuristic with a definitive per-file signal: await the actual write (wrap the plugin's write, or write the `.d.ts` ourselves from the compiled output).
- Propagate compile errors out of `transform` so a broken module throws deterministically.

packages/components/dev/scss-types/generateScssTypes.ts
```diff
@@ -111,12 +101,16 @@ const run = async (): Promise<void> => {
...
   await configResolved(config);
+
+  // Sequential: only the in-flight declaration is ever deleted, so a failure
+  // leaves every other committed file untouched.
   for (const scssFile of scssFiles) {
+    const dtsFile = toDtsPath(scssFile);
+    rmSync(dtsFile, { force: true });
     await transform("", scssFile);
+    await waitForRewrite(dtsFile, scssFile);
   }
```
mfal, line 113: Await `Promise.all(scssFiles.map(async(file) => ...))`?

mfal (approved):

PR #2819, opened 2026-08-07
fix(components): keep the global stylesheet out of tree shaking

Lisa18289:
Every Storybook preview — PR previews and https://storybook.flow-components.de alike — renders without design tokens, reset or fonts.

## Cause

`packages/components/package.json` declares

```json
"sideEffects": ["**/*.css", "**/*.scss"]
```

since #2772. That marks every `.ts`/`.tsx` file in the package side-effect free. `src/styles/index.ts` is exactly such a file — its only job is `import "./index.scss"`. Rollup drops it, and the design tokens, the reset and the fonts go with it.

`.storybook/preview.tsx` imports `../src/styles`, so this hits Storybook: the built preview contains no `:root` token definitions in any of its 95 CSS chunks, and `iframe.html` links no stylesheet at all.

The library build is unaffected. It runs in lib mode with `preserveModules`, so `dist/css/all.css` still carries the tokens — consumers of the published package see nothing of this.

## Fix

Add both side-effect-only entrypoints to the allowlist: the source path for in-repo builds (Storybook), the emitted `dist/js/globals.mjs` for the published package. Consumer tree shaking stays intact.

## Verification

`storybook build` before the change: no `:root` anywhere in `storybook-static/assets/*.css`, no `<link rel="stylesheet">` in `iframe.html`.

After: `iframe.html` links `assets/iframe-*.css`, which contains the token definitions (`--corner-radius--round:`) and the reset (`all:initial`).

The same absence is observable in the deployed artifacts — PR preview and production Storybook ship identical chunk hashes, neither with tokens.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -7,7 +7,9 @@
   "repository": "https://github.com/mittwald/flow",
   "sideEffects": [
     "**/*.css",
-    "**/*.scss"
+    "**/*.scss",
+    "./src/styles/index.ts",
+    "./dist/js/globals.mjs"
   ],
   "exports": {
     ".": {
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -8,8 +8,7 @@
   "sideEffects": [
     "**/*.css",
     "**/*.scss",
-    "./src/styles/index.ts",
-    "./dist/js/globals.mjs"
+    "./src/styles/index.ts"
   ],
   "exports": {
     ".": {


packages/components/package.json
```diff
@@ -7,7 +7,9 @@
   "repository": "https://github.com/mittwald/flow",
   "sideEffects": [
     "**/*.css",
-    "**/*.scss"
+    "**/*.scss",
+    "./src/styles/index.ts",
+    "./dist/js/globals.mjs"
```
ins0: dist files are not correct in sideEffects, styles/index.ts is correct

Lisa18289: Agreed — removed in d8467441a. `dist/js/globals.mjs` has no `exports` entry, so nothing can import it; the source path is what the in-repo builds resolve and it is enough.

Re-verified with `storybook build`: `iframe.html` links `assets/iframe-CzTs1BpE.css`, which carries the token definitions and the reset.

ins0 (approved):

PR #2820, opened 2026-08-07
fix(Autocomplete): prevent the options popover from collapsing

Lisa18289:
Most [Autocomplete examples in the docs](https://flow.mittwald.de/04-components/form-controls/autocomplete/overview) showed no results while typing.

## Why

The popover was placed correctly, right below the field — it just had `max-height: 0`, so the options were in the DOM but the box had no height.

The popover is portaled into the Autocomplete itself (so `useFocusWithin` covers it too, #1738). The container had no positioning context, so the popover's containing block was the next positioned ancestor — in the docs `div.mainWrapper`, near the top of the page. react-aria's `getMaxHeight` then mixes coordinate systems: `boundingRect.bottom` forgets `boundaryToContainerTransformOffset` in its `visualViewport` term, and `Math.min` picks exactly that one. Once a field sits more than one viewport below the ancestor, the available height comes out as 0. Opening *upwards* was fine: that path uses `boundingRect.top`, which is computed consistently.

## Fix

- The Autocomplete container gets `position: relative` and is thereby the popover's containing block, so the offset stays small.
- react-aria measures its container padding from that container, so the default 16px would offset the options from the input — the options popover passes `containerPadding={0}`. `PopoverContent` accepts the prop for that; default stays 16.
- Docs: `.liveCodeEditor` no longer clips, because the box is only as tall as the example inside it. `.actions` rounds its own bottom corners instead.

## Verified

New browser test reproduces the bug without the fix (`max-height: 0px`, option not hit-testable) and catches the 17px misalignment without `containerPadding={0}`. Both fixes also checked against the live docs markup: before, all four options clipped; after, all visible. Existing visual baselines unchanged.

## Follow-up

The actual defect is upstream in react-aria. A pnpm patch would only fix this repo, not mStudio or extensions — worth reporting separately.


packages/components/src/components/Popover/components/PopoverContent/PopoverContent.tsx
```diff
@@ -34,7 +41,7 @@ export const PopoverContent: FC<PopoverContentProps> = (props) => {
     <Aria.Popover
       {...rest}
       className={className}
-      containerPadding={16}
+      containerPadding={containerPadding}
```
ins0: we should fix this over css and not introduce a new public property that differs from the token-style visual approach

ins0 (approved):

PR #2828, opened 2026-08-10
feat(Link): support Button in Link

### LINKED ISSUE #2360: Button vs Link
We have repeatedly received questions about the strict distinction between Button and Link.

Contributors often want to use a call to action for navigation (for example opening an external page or subscribing to something) and are unsure why this must be implemented as a link instead of a button.

Examples of questions that came up:

> "Ich möchte doch aber explizit zu einer Aktion aufrufen, nämlich das GitHub Issue zu abonnieren. Warum muss das dann ein Link sein und keine Call2Action?"

> "Dann machen wir da einfach einen Button rein, mit 'Extension [extern] aufrufen' [...] Ja, aber warum nicht? Das ist ja die Hauptaktion."

Lisa18289:


apps/docs/src/content/04-components/navigation/link/overview.mdx
```diff
@@ -33,13 +33,20 @@ HSL Lightness-Wert von mehr als 50.
...
 ---
 
-# Kombiniere mit ...
+# States
+
+Ein Link hat vier verschiedene States: **Default**, **Hover**, **Pressed** und
+**Disabled**.
 
-## Icon
+<LiveCodeEditor example="disabled" editorCollapsed />
```
ins0, line 40: the example should display the annotation above it, so all states should be illustrated in this example and not `disabled` only.

packages/components/src/components/Button/Button.tsx
```diff
@@ -185,6 +183,34 @@ export const Button = flowComponent("Button", (props) => {
...
       </PropsContextProvider>
       {stateIcon}
+    </>
+  );
+
+  if (elementType === "span") {
+    const spanProps = filterDOMProps(restProps, { global: true });
+
+    return (
+      <span
```
ins0, line 193: since we return a new dom element we should check if the `render` prop from `Aria.Button` is a better alternative to render a span here - but within the aria lib.

Lisa18289: react aria does not support this:

> You must render the expected element type (e.g. if `<button>` is expected, you cannot render an `<a>`).

i still tried, but it resultet in a very broken span, with tabindex and button props that should not be spread into the span element

Lisa18289: it works now, but  React Aria warns in dev because the element type no longer matches what it expects.

ins0 (approved):

PR #2829, opened 2026-08-11
docs(homepage): let the developer experience example compose itself

### LINKED ISSUE #2824: docs(homepage): "Fokus auf Developer Experience" example doesn't demonstrate its own copy
The homepage tile **"Fokus auf Developer Experience"** has a live code example that doesn't show what the copy describes.

## Problem

The tile text claims (`apps/docs/src/app/page.tsx:105`):

> Verschachtelte Components erleichtern das Einfügen. Automatisierungen ordnen Components automatisch mit passenden Abständen an und verhindern Abweichungen vom Design System.

But the live example (`apps/docs/src/app/page.tsx:43`) is a single, plain button:

```tsx
import { Button } from "@mittwald/flow-react-components";

<Button color="primary">
  Button
</Button>
```

None of the three points the copy makes are visible:

- **"Verschachtelte Components"** — nothing is nested; it's one leaf component.
- **"Automatisierungen ordnen Components ... mit passenden Abständen an"** — there's no composition, so no automatic spacing/arrangement to see.
- **"verhindern Abweichungen vom Design System"** — not demonstrated either.

The example undersells exactly the DX strengths the tile is meant to showcase.

## Suggested fix

Replace the example with one that actually composes nested Flow components and lets the automatic layout/spacing do the work — e.g. a `Section` with a `Heading` + `Text` + an `ActionGroup`/`ButtonGroup`, or a small `LayoutCard` composition. The point is to show that nesting components produces a well-spaced, on-brand result without manual styling.

The "Zu den Components" link already points at `action-group` — an action-group-based example would tie the copy, the link, and the demo together.

**File:** `apps/docs/src/app/page.tsx` (`code` constant, ~line 43)

Lisa18289:
Closes #2824

The "Fokus auf Developer Experience" tile claimed nested components, automatic spacing and design system conformance while showing a single plain `Button`. None of the three was visible.

The example is now a profile section that composes itself from the bare `<Section>` — Header with its heading, then the labeled value, then the action. Every added layer snaps into place with the right spacing, and the header pushes its action to the opposite end on its own. That also ties the tile to its "Zu den Components" link.

```tsx
<Section>
  <Header>
    <Heading>Profil</Heading>
    <Button color="secondary" variant="soft">
      Bearbeiten
    </Button>
  </Header>
  <LabeledValue>
    <Label>E-Mail-Adresse</Label>
    <Content>max@mustermann.de</Content>
  </LabeledValue>
</Section>
```

The editor shows the composition without its import — spelled out, seven component names take more room than the example itself. The editor scope comes from a real import statement in `codeSteps.ts`, just one the reader never sees.

## How it behaves

- **Starts when the tile scrolls into view** (IntersectionObserver, 30 %). The tile sits below the fold; otherwise the animation would be over before anyone sees it.
- **Nothing moves.** The finished example is rendered server side; preview and code block each measure their final height on mount and hold it while typing, and the preview is top aligned during the run.
- **`prefers-reduced-motion`**: the last step immediately, no animation.
- **Stays a live example.** Focusing or clicking the editor jumps to the end and stops the animation; editing works as before.
- **`transformCode` keeps the last parsable preview** instead of rendering a parser error while the code is half typed. This also applies to manual edits.

## Implementation

`page.tsx` no longer builds the `LiveProvider` inline — it renders `<HomeCodeExample />`, which keeps the animation's re-renders out of the page component.

- `codeSteps.ts` — the four steps plus the import statement the scope is derived from. Each step inserts one block at a single position, inside `<Header>` or `<Section>`. That is what makes typing the difference between two steps possible.
- `useTypedCode.ts` — describes step *n+1* as one insertion into step *n*, expands it into frames and drives them with a timer chain. Insertions are normalised to whole lines, so a new block is typed on its own line instead of growing out of the line below it.
- `HomeCodeExample.tsx` — editor, height reservation, visibility trigger.

## Verified

Driven with Playwright against `pnpm nx dev docs`, sampled every 400 ms across the whole run:

- Tile 430 px, preview 143 px, code block 285 px, first code line 161 px — every value constant, and the heading stays at its offset of 25 px from the moment it appears.
- No horizontal scrolling in the code block at 1280 px, 1100 px or 390 px.
- Off screen the finished example stays put; scrolling in restarts it at `<Section></Section>` and it completes.
- Reduced motion keeps the complete example, on load and on scroll.
- Clicking the editor mid-run jumps to the end; editing "Bearbeiten" afterwards updates the button in the preview.
- Dark theme and 390 px width (no horizontal page overflow).

`pnpm lint`, `prettier --check` and `tsc --noEmit` pass.

### DIFF
--- a/apps/docs/src/app/_components/HomeCodeExample/HomeCodeExample.tsx
+++ b/apps/docs/src/app/_components/HomeCodeExample/HomeCodeExample.tsx
@@ -1,0 +1,109 @@
+"use client";
+import type { FC } from "react";
+import {
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from "react";
+import {
+  LiveEditor,
+  LivePreview,
+  LiveProvider,
+} from "@mfalkenberg/react-live-ssr";
+import styles from "@/lib/liveCode/components/LiveCodeEditor/LiveCodeEditor.module.css";
+import { flowTheme } from "@/lib/liveCode/components/LiveCodeEditor/lib/flowTheme";
+import extractDefaultExport from "@/lib/liveCode/components/LiveCodeEditor/lib/extractDefaultExport";
+import { extractEditorScope } from "@/lib/liveCode/components/LiveCodeEditor/lib/extractEditorScope";
+import { codeSteps } from "./codeSteps";
+import { useTypedCode } from "./useTypedCode";
+
+const finalCode = codeSteps.at(-1) ?? "";
+const scope = extractEditorScope(finalCode);
+
+const useIsomorphicLayoutEffect =
+  typeof window === "undefined" ? useEffect : useLayoutEffect;
+
+/**
+ * The live example of the "Fokus auf Developer Experience" tile: it composes
+ * itself step by step, so the nesting and the automatic spacing the tile talks
+ * about happen in front of the reader. Stays a fully editable live example
+ * afterwards.
+ */
+const HomeCodeExample: FC = () => {
+  const editorRef = useRef<HTMLDivElement>(null);
+  const [reservedHeight, setReservedHeight] = useState<number>();
+  const [isVisible, setIsVisible] = useState(false);
+
+  // The example starts out complete: its final height is measured up front and
+  // held while the animation runs, so growing code never pushes the page
+  // around. The animation itself waits for the tile to be on screen — the tile
+  // sits below the fold and watching it compose itself is the whole point.
+  useIsomorphicLayoutEffect(() => {
+    const element = editorRef.current;
+
+    if (element === null) {
+      return;
+    }
+
+    setReservedHeight(element.getBoundingClientRect().height);
+
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        if (entry?.isIntersecting) {
+          setIsVisible(true);
+          observer.disconnect();
+        }
+      },
+      { threshold: 0.3 },
+    );
+
+    observer.observe(element);
+
+    return () => observer.disconnect();
+  }, []);
+
+  const { code, isTyping, skip } = useTypedCode(
+    codeSteps,
+    isVisible && reservedHeight !== undefined,
+  );
+
+  const lastPreview = useRef("");
+
+  const transformCode = useCallback((code: string) => {
+    try {
+      lastPreview.current = extractDefaultExport(code);
+    } catch {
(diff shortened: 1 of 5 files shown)


mfal:
Two ideas from a second pass at #2824 that I'd like reflected here.

## 1. Start from loose components, not from the wrapper

The composition currently grows out of a bare `<Section>` — the wrapper is on screen before there is anything for it to arrange. The stronger "before" picture is no wrapper at all: the components as siblings in a fragment, rendered one below the other in the order they are written. The nesting step then *is* the moment the automation becomes visible.

```
1. <>          <Heading> and <Button>, stacked
2. <>          + <Label> + <Content>, stacked
3. <>          <LabeledValue> ties the label to its value
4. <Section>   <Header> lifts the button up next to the heading
```

The fragment doubles as the placeholder `<Section>` replaces, so the last step is one visible swap. That turns "Automatisierungen ordnen Components automatisch an" into something you watch happen instead of something the reader has to take on faith.

## 2. Cycle through more than one example

One example is over after a few seconds and the tile is static again. Three, each with the same beat — loose, then nested — show that the rule holds everywhere, not just in a `Section`. Each finished example holds ~10s before the next starts:

- the profile section from this PR
- `<Combine>`: `<Initials>` alone → wrapped in `<Avatar>` → next to a `<Text>`
- `<IllustratedMessage>`: `<IconApp />` → + `<Heading>` → + `<Text>` → + `<Button>` → the whole group centered into an empty state

## Two mechanics, in case they are useful

- **Keep half-typed code out of the preview by construction.** Instead of `transformCode` falling back to the last parsable result, drive the two from separate strings: `<LiveEditor code={halfTypedFrame} onChange={…} />` gets the frame, `<LiveProvider code={lastCompletedStep}>` gets the preview. `LiveEditor` spreads its own props over the context values, so both overrides work. A parse error from a real edit still surfaces.
- **Constant height without measuring.** Pad every frame with blank lines up to the line count of the longest step of any example. The code block then has its final height from the first frame, with no on-mount measurement — only the preview needs a `min-height`.

Related: with `prefers-reduced-motion` it is worth subscribing to the media query rather than reading it once on mount, so flipping the setting on an open page stops the loop.

mfal:
Opened #2901 as an alternative — same problem, the ideas from the comment above implemented: the example starts without a wrapper and nests itself, and three examples run in sequence (profile section, `Combine`, empty state). Worth comparing the two and picking one.

Lisa18289:
better example: https://github.com/mittwald/flow/pull/2901

PR #2843, opened 2026-08-12
feat(docs): allow switching off component grouping

### LINKED ISSUE #2801: Docs: Gruppierung in Component Navigation deaktivieren
### Project Information

mStudio extension, custom project

### Feedback

Ich habe öfters das Problem, dass ich weiß, wie eine Component heißt und da in der Doku was nachschauen möchte, aber dann nicht weiß, in welcher Gruppe der Component-Liste die Component eingeordnet ist. In dem Fall würde es mir helfen, die Components ohne Gruppierung alphabetisch sortiert aufgelistet zu haben.

**Feature-Wunsch:** Gruppierung der Component-Liste (Menü links) deaktivieren

Für mich sind das zwei Unterschiedliche UseCases:
* Mit Gruppierung: Exploration, rausfinden was es gibt, wo was geht („welche Form-Elemente gibt es?“ „welche Möglichkeiten zur Strukturierung von Inhalt habe ich?“)
* Ohne Gruppierung: Ich weiß schon spezifisch nach welcher Component ich suche und möchte die möglichst schnell finden

Beide Varianten haben also ihre Berechtigung und ich würde gerne einfach zwischen beiden umschalten können.

Lisa18289:
Closes #2801

## What

- The component navigation is split into two top-level sections: **Components** (all groups except React Hook Form) and **Integrations** (the React Hook Form navigation group), in that order.
- A settings context menu (`IconAdjustments`) next to the "Components" heading holds a switch **Gruppiert**. On (default) keeps the category grouping, off shows one flat alphabetical list — in the navigation *and* in the components overview.
- The preference is persisted per user in `localStorage` and shared by both views through a small `useSyncExternalStore` hook.

## Notes

- **Why localStorage and not a cookie like the theme:** the docs are a static export (`output: "export"`), so nothing reads a cookie before the first paint. And unlike the theme — a single CSS attribute an inline script can set — grouping changes the DOM structure, so the inline-script trick does not apply. Consequence: on load the default (grouped) renders first and switches after hydration.
- **Alphabetical mode sorts purely by name** — deprecated components are no longer pushed to the end of the list (grouped mode keeps that behaviour). Their status badge stays visible.
- The components overview page keeps its category grouping; the Components/Integrations split is navigation-only, as discussed.
- Verified in the dev server: toggle in both directions, persistence across reload, component detail pages, other doc sections unchanged, mobile off-canvas navigation. `eslint`, `prettier` and `tsc --noEmit` are clean.

### DIFF
--- a/apps/docs/src/app/04-components/_components/ComponentsList.tsx
+++ b/apps/docs/src/app/04-components/_components/ComponentsList.tsx
@@ -14,6 +14,7 @@ export interface ComponentLink {
 
 interface Props {
   components: ComponentLink[];
+  "aria-label"?: string;
 }
 
 export const ComponentsList: FC<Props> = (props) => {
@@ -23,6 +24,7 @@ export const ComponentsList: FC<Props> = (props) => {
 
   return (
     <List.List
+      aria-label={props["aria-label"]}
       getItemId={(component) => component.id}
       defaultViewMode="tiles"
       batchSize={components.length}
--- a/apps/docs/src/app/04-components/_components/ComponentsOverview.tsx
+++ b/apps/docs/src/app/04-components/_components/ComponentsOverview.tsx
@@ -1,10 +1,9 @@
-"use client";
 import type { FC } from "react";
 import { Heading, Section } from "@mittwald/flow-react-components";
 import { groupBy } from "remeda";
 import type { ComponentLink } from "@/app/04-components/_components/ComponentsList";
 import { ComponentsList } from "@/app/04-components/_components/ComponentsList";
-import { useComponentGrouping } from "@/app/_lib/componentGrouping";
+import { ComponentGroupingView } from "@/app/_components/ComponentGroupingView/ComponentGroupingView";
 import { extractTextFromPath } from "@/app/_lib/extractTextFromPath";
 
 interface Props {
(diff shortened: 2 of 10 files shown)


maaaathis (changes requested):
When loading the page, the grouped view briefly flashes before switching to the alphabetical view, when the view is configured to be alphabetical. Looks like a hydration error.

https://github.com/user-attachments/assets/c4dbe73e-10ae-4540-9e02-10ad0184baeb

Lisa18289 (commented):
Fixed in 04a8cfe18.

The flash came from React only learning the preference after hydration, so the prerendered HTML always showed the grouped default. The decision now happens before the first paint, without React:

- Both views are always rendered, and CSS picks one via a `data-component-grouping` attribute on `<html>`.
- A blocking inline script in `<head>` sets that attribute from localStorage while the head is parsed. Measured on the static export (`nx build docs`, served locally): script at 23ms, first paint at 88ms — and a MutationObserver confirms hydration does not strip the attribute again.
- Toggling the switch flips the attribute, so it no longer re-renders the navigation or the list either.

Cost of the second view: ~20KB extra markup for the navigation, and the components overview roughly doubles its card markup. If that weight is a concern, the alternative would be prerendering the ungrouped overview as its own route — happy to switch if you prefer that.

Note it does not use Next's `<Script strategy="beforeInteractive">` (as the theme does): in the static export that ends up in Next's deferred script queue, which can run after the first paint. A raw inline `<script>` is what guarantees the ordering.

maaaathis (approved):

maaaathis (approved):

LukasFritzeDev:
❤️ love it

Issue #2883, opened 2026-08-25
Breaking changes since alpha.600 are missing from the changelog and the migration guide


mfal:
Every breaking change released since `0.2.0-alpha.600` is missing from the changelog, and five of seven have no migration note either. A consumer upgrading to the latest version cannot find out what they have to change.

Found while re-verifying an mStudio extension developer's upgrade report (7 extension repos, `alpha.621`…`alpha.715` → `alpha.1018` → today).

## The breaking changes at `>=0.2.0-alpha.600`

Every commit carrying the conventional-commit `!` marker, mapped to the release it shipped in:

| Release | Change | MIGRATION.md | CHANGELOG |
| --- | --- | --- | --- |
| `alpha.646` | `Action`: `action` prop removed from `ActionProps` (`9e0ff4a26`, #2255) | ✗ | ✗ |
| `alpha.694` | Forms in Modals auto-reset on close (`99d596209`) | ✗ | ✗ |
| `alpha.712` | `MutedActionError` renamed to `AbortActionError` (`9ac0a4a95`) | ✗ | ✗ |
| `alpha.786` | `AccentBox`: `color` split into `backgroundColor` + `color` (`be602ba7a`, #2434) | ✓ | ✗ |
| `alpha.802` | `mittwald-password-tools-js`: `AsyncRule`/`SyncRule` dropped for `Rule` (`1483bd7f6`, #2461) | ✗ | ✗ |
| `alpha.913` | `PasswordCreationField` moved off the default export surface (`5bb91603a`, #2646) — reverted in `alpha.1000` (`42095b5e7`) | n/a | ✗ |
| `alpha.1046` | `Button`: color `accent` renamed to `success` (`34ad687ae`, #2874) | ✓ | ✗ |

2 of 7 in the migration guide, 0 of 7 in any changelog.

## 1. The `!` marker suppresses the changelog entry

`grep -c BREAKING` on `CHANGELOG.md` and `packages/components/CHANGELOG.md` returns `0`. None of the seven commits above appears in either file.

Release `alpha.1046` contains exactly one non-release commit — `feat(Button)!: rename color "accent" to "success"` — and both changelogs say "Version bump only". The structurally identical `feat(Combine): rename Align to Combine` one release later (`alpha.1047`, no `!`) is listed normally.

[...]

PR #2884, opened 2026-08-25
docs(migration): add missing migration notes and the Action codemod

mfal:
Closes the documentation half of #2883: the breaking changes released since `0.2.0-alpha.600` that shipped without a migration note, plus the missing link from the docs site.

## Migration notes

| Release | Change | What was missing |
| --- | --- | --- |
| `alpha.646` | `Action`: `action` renamed to `onAction` | note + codemod |
| `alpha.694` | `Form` resets after the surrounding modal closes | note |
| `alpha.712` | `MutedActionError` → `AbortActionError` | note |
| `alpha.802` | password-tools: `AsyncRule`/`SyncRule` → `Rule` | note |
| `alpha.1021` | ext-bridge: `HostConfig` no longer imported from the unpublished `@mittwald/flow-core` | note |

The `alpha.646` entry joins the existing `alpha.637` → `alpha.646` section, which covered the ResetButton/SubmitButton interfaces from the same release but not `Action`.

`alpha.913` (`PasswordCreationField` off the default export surface, #2646) needs no entry — it was reverted in `alpha.1000`.

## New codemod

`flowAlphaActionPropToOnAction` renames the prop on `Action`. `action` was removed from `ActionProps` while the runtime still maps it to `onAction`, so consumers hit a `tsc` error with no `@deprecated` marker and nothing to search for.

Scoped like the existing codemods: only elements resolving to `Action` imported from `@mittwald/flow-react-components` or `@mittwald/flow-remote-react-components` (including subpath entries). Verified against fixtures covering named imports, `as` aliases, `import * as Flow` namespaces, a same-named `Action` from another package, a plain `<form action="…">`, and an element that already has `onAction` — that one keeps `onAction` and only loses the stale `action`, mirroring the runtime fallback.

## Discoverability

The versioning page explained the policy but linked neither migration guide nor the codemods, so `MIGRATION.md` was reachable only from `README.md` and, inline, from the Align component page. It now has a **Migration bei einer Breaking Change** section: both migration guides, how to run a codemod, and how to collect deprecation warnings with `DeprecationWarningProvider`.

## Removed

`flowAlphaPasswordCreationField` belongs to the reverted #2646. The revert removed its MIGRATION.md entry but left the transform, which rewrites imports to a `password-tools` subpath `packages/components/package.json` does not export — the entry is `./mittwald-password-tools-js`. Since the versioning page now links the transforms folder, the folder should not contain a transform that breaks a consumer's imports.

## Still open from #2883

- **Item 1, the systemic one:** `!:`-marked commits never appear in any changelog (`grep -c BREAKING` → 0 across both files). Until that is fixed, these notes stay invisible to anyone upgrading via the changelog.
- **Item 5:** the nine live deprecations are not yet collected as a 2.0.0 removal list.
- Unrelated to #2883, noticed while typechecking: `packages/codemods` has no `test:compile` script, and `flow020.ts` has a standing `tsc` error nothing catches.


mfal:
Added `flowAlphaAll` — one codemod that runs every `0.2.0-alpha` migration transform in release order, so catching up from an older alpha is a single invocation instead of four.

It is a jscodeshift transform rather than a shell script on purpose: `@mittwald/flow-codemods` is private, so a script would need a new distribution path (and `curl | sh` is not one I want to document). A transform reuses the exact `npx jscodeshift -t <raw url>` call consumers already know, and jscodeshift keeps doing traversal, parsing and parallelism.

```shell
npx jscodeshift \
  -t https://raw.githubusercontent.com/mittwald/flow/refs/heads/main/packages/codemods/src/transforms/flowAlphaAll.ts \
  --parser tsx \
  src
```

## Two transforms are deliberately excluded

Both would do damage on a current codebase, so "all" means "all migration transforms":

- **`flowRemote`** ports an app to `@mittwald/flow-remote-react-components`. That is a port to a different package, not a migration — running it on a normal app rewrites every Flow import.
- **`flow020`** migrates `0.1.0` → `0.2.0` by collapsing subpath imports onto the package root. It special-cases only `doc-properties`, `all.css`, `react-hook-form` and `nextjs`; everything else goes to the root. The package has since gained `/internal`, `/flr-universal` and `/mittwald-password-tools-js`, so on a current codebase it would silently break those imports. It stays a separate, opt-in step, and both `MIGRATION.md` and the transform's doc comment say so.

## Verified

One fixture combining all four changes in a single file, one fixture that must stay untouched:

| Input | Output |
| --- | --- |
| `Align` import + both JSX tags | `Combine` |
| `<Heading color="primary">`, `<Link color="primary">` | `color="default"` |
| `<Button color="primary">` | unchanged — still valid |
| `<Action action={save}>` | `onAction={save}` |
| `<Button color="accent">` | `color="success"` |
| `Action`/`Link` from another package, `<form action="/submit">` | unchanged |

ESLint and Prettier clean; `tsc --noEmit` reports nothing for the new file (the two standing errors in `flow020.ts` and the now-deleted `flowAlphaPasswordCreationField.ts` are unrelated, and `packages/codemods` has no `test:compile` script that would have caught them).

Lisa18289 (approved):

PR #2889, opened 2026-08-25
feat(ExtBridge): declare context parameters and align config type with runtime

### LINKED ISSUE #2885: ext-bridge config: type strips unknown keys, runtime keeps them
The ext-bridge config schema and its runtime parsing disagree about unknown keys. The published type strips them, the runtime keeps them. Extensions that rely on host-supplied extra keys have to narrow at runtime (`typeof x === "string"`) because the type says the key does not exist.

Reported by an mStudio extension developer who hit it across seven extension repos while upgrading to `0.2.0-alpha.1018`. Two extensions read a host-supplied extra key: `containerId` (extension-webssh) and `mailAddressId` (extension-mobileconfiggenerator).

## The contradiction

`packages/ext-bridge/src/config/schemas.ts` exports the schema without a catchall, so `z.infer` — and with it `ExtBridgeConfig`, what `useConfig()` and `getConfig()` return — has no index signature:

```ts
export const config = baseConfig.extend(contextParameters.shape);
```

`packages/ext-bridge/src/config/parse.ts` adds the catchall back for the actual parse:

```ts
const parsed = config.catchall(z.string().optional().nullable()).safeParse(something);
```

The seam is `readiness.ts`:

```ts
const parsedConfig = parseConfig(config);      // wide: carries unknown keys
const hostConfig = extractHostConfig(config);
mwExtBridge.config = { ...parsedConfig, ...hostConfig };   // narrowed to ExtBridgeConfig
```

`mwExtBridge.config` is typed `ExtBridgeConfig`, so the extra keys survive the assignment at runtime and vanish from the type.

Confirmed in the published `0.2.0-alpha.1049` artifacts: `dist/types/config/schemas.d.ts` ends in `z.core.$strip`, `dist/js/config/parse.mjs` calls `.catchall(...)`.

## Since when

Commit `5a8fe9a29` (`fix(Remote): fix config types`), released in `0.2.0-alpha.790`. It moved the `.catchall()` off the exported schema and into `parseConfig`. Verified against the published tarballs: `alpha.789` types say `z.core.$catchall`, `alpha.790` say `z.core.$strip`.

The same commit deleted the type test that pinned the old contract:

```ts
function testUnknownContextParameters() {
  expectTypeOf<string | null | undefined>(testConfig.unknownProp);
}
```

So the guard was removed rather than the behaviour aligned, and nothing has failed since.

## The host side has the same hole

`RemoteExtBridgeConfig` in `packages/remote-core/src/connection/types.ts` is `Omit<ExtBridgeConfigInput, keyof HostConfig>`. With `$strip` in the type, an mStudio host that wants to pass `containerId` has no typed way to declare it either. Whatever we do should cover both directions: the host writing an extra key and the extension reading it.

Note the runtime catchall is load-bearing. zod's default `strip` deletes unknown keys from the parsed output — verified with the repo's zod 4.4.3:

```
strip   : {"sessionId":"s"}
catchall: {"sessionId":"s","containerId":"c"}
```

Without the `.catchall()` in `parseConfig`, `containerId` and `mailAddressId` would be dropped, not just untyped.

### A — put the catchall back on the exported schema

One line: move `.catchall(z.string().optional().nullable())` back into `schemas.ts` and restore the deleted type test. Type and runtime agree again, no runtime change, no breaking change.

The cost is what `5a8fe9a29` presumably fixed: the catchall's index signature makes every property access type-check. `config.projectIdd` then resolves to `string | null | undefined` instead of erroring, and `RemoteExtBridgeConfig` accepts any object on the host side. That is a real regression in typo protection, which is why this is the fallback rather than the recommendation.

### B — drop the runtime catchall

Makes the runtime match the current type. Rejected: it silently deletes keys that extensions read today, so webssh and mobileconfiggenerator break with no type error to warn them. If we ever want this, it needs a deprecation cycle, and it still leaves extensions without a way to receive extra keys.

### C — a declared extension point (recommended)

Keep the runtime catchall and let a consumer declare the extra keys it expects, so only that consumer's config type widens:

```ts
const config = useConfig<{ containerId: string }>();
config.containerId; // string | undefined
```

`getConfig` and `useConfig` take an optional type parameter and return `ExtBridgeConfig & Partial<Extra>`. Additive, no breaking change, no index signature for anyone who does not opt in, and typos stay errors. The symmetric parameter on `RemoteExtBridgeConfig` gives the host a typed way to supply the key.

A declaration-merging variant (an `ExtBridgeConfigExtensions` interface the extension augments once) is the alternative if we want the keys declared per project instead of per call site. It reads better at the call sites and worse in the type errors.

Whichever we pick, restore a type test in `schemas.test-types.ts` so the contract is pinned again.

## Recommendation

C with the type parameter, plus the restored type test. Fall back to A if we want the contradiction gone in one line and accept losing typo protection. Not B without a deprecation cycle.

Either way the extensions can drop their runtime narrowing, which is the remaining workaround from the upgrade report — see #2883 for the documentation half of the same report.

mfal:
## What & why

Closes #2885.

The ext-bridge config schema and its runtime parse disagreed about unknown keys:
the exported schema stripped them, `parseConfig` added a `.catchall()` back for
the actual parse. So an extension reading a host-supplied context parameter —
`containerId` in extension-webssh, `mailAddressId` in
extension-mobileconfiggenerator — had to narrow at runtime
(`typeof x === "string"`) because the type said the key did not exist. In place
since alpha.790, which also deleted the type test that pinned the old contract.

Implements **option A** with the known-parameter list, as decided in the issue.

### Changed

- **26 context parameters declared** in `contextParameters`
  (`packages/ext-bridge/src/config/schemas.ts`). `emailAddressId` and
  `ingressId` are deprecated in favour of `mailAddressId` / `domainId`, and stay
  declared.
- **`.catchall()` back on the exported schema**, so type and runtime agree and a
  parameter mStudio starts sending before Flow declares it still arrives. The
  accepted cost is the index signature: a typo resolves to `string | undefined`
  instead of erroring.
- **One shared schema for every optional config value**, typed
  `string | undefined` and mapping `null` → `undefined`
  (`z.string().nullable().optional().transform(…)`). A host that still sends
  `null` degrades to "value absent" instead of failing the parse — a failed
  parse throws `ExtBridgeError` and the extension never starts.
- **Type tests restored and extended** in `schemas.test-types.ts`: the unknown-key
  contract, the declared and deprecated parameters, and that the required base
  properties are not collapsed into the index signature.
- **`RemoteExtBridgeConfig`** now points at the new
  `ExtBridgeConfigInputWithoutHostConfig` instead of
  `Omit<ExtBridgeConfigInput, keyof HostConfig>` — `Omit` over a type with an
  index signature collapses into that index signature alone and drops every
  declared property, which broke `getWithMergedHostConfig` outright.
- **Migration notes** in `packages/ext-bridge/MIGRATION.md`, backfilled since
  alpha.600 (remote side only): this change, Node 24 (alpha.941), the alpha.790
  type regression, and the React peer range (alpha.647).
- **`.claude/templates/release-notes.md`** gets the rule that keeps the host side
  out of drafted release notes and migrations. It came up while writing the
  notes for this change: the host contract ships with mStudio and never
  migrates, and package boundaries do not decide it — ext-bridge is
  extension-facing yet carries host types. The rule says to judge per symbol and
  names symbols only as examples, so it cannot go stale.

### Two decisions worth a reviewer's eye

**The `@deprecated` tags live in `config/types.ts`, not on the zod shape.** JSDoc
on a zod shape does not survive the declaration emit — verified against the built
artifact: a consumer importing from `dist/types` got no deprecation at all. With
`DeprecatedContextParameters` intersected into `ExtBridgeConfig`, TypeScript
reports `'ingressId' is deprecated.` / `'emailAddressId' is deprecated.`, so the
strikethrough and the successor show up at the call site. `schemas.ts` carries a
pointer comment so the next maintainer finds them.

**`null` is tolerated, not rejected.** The issue's decision line said bare
`.optional()`; the transform reaches the same public type without turning a stray
`null` from a host we do not control into a boot failure.

### Type-breaking

Optional config values lost `null` from their type — `variantKey` was
`string | null | undefined`, now `string | undefined`, and the same for every
context parameter and unknown key. `typeof x === "string"` and truthiness checks
keep working; only an explicit `=== null` branch becomes dead code. No codemod:
deleting such a branch is a semantic decision. Covered by the MIGRATION.md entry
per the type-break rule on the versioning page.

## Verification

- `pnpm nx run-many -t test:compile` — 8 projects green
- `pnpm nx run-many -t test:unit` — 212 tests green
- `pnpm lint` — clean (pre-existing warnings only)
- Runtime behaviour checked against the **built** `dist/js/config/parse.mjs`:
  known + undeclared keys survive, `null` becomes "absent", a non-string still
  throws `ExtBridgeError`
- Emitted types checked as a consumer would see them: `dist/types` now ends in
  `z.core.$catchall` (was `$strip`), `containerId` and unknown keys are
  `string | undefined`, `sessionId` stays required, both deprecated parameters
  report their replacement

No rendered behaviour changed, so no visual snapshots are involved.

## Checklist

- [x] `pnpm lint` is clean and the compile/unit suites pass
- [x] **Generated code is committed** (no generators affected)
- [x] User-facing strings — none
- [x] Docs: the documentation half of the report is tracked separately in #2883


Lisa18289 (approved):

mfal:

Lisa18289 (approved):

PR #2900, opened 2026-08-25
ci(dependabot): group every npm update and auto-merge on green CI

mfal:
Dependency updates had stopped landing. Since 2026-07-24 not a single `dev-dependencies` group PR merged — it was recreated every week and grew 12 → 16 → 17 → 23 → 24 → **29 updates** — while ~8 ungrouped production PRs opened alongside it every Monday.

The mechanism is quadratic, not accidental: each merge invalidates `pnpm-lock.yaml` in every other open PR, Dependabot rebases them, CI runs again. With ~10 open PRs that cascade never drains — and each one still needs a code owner approval by hand.

## What changes

**Four npm PRs a week instead of ten.** Every npm dependency falls into a group; nothing is left ungrouped:

| Group | Contents | Why separate |
| --- | --- | --- |
| `react-aria` | `react-aria*`, `@react-aria/*`, `@react-stately/*`, `@react-types/*`, `@internationalized/*` | Moves as one unit, breaks as one unit — a partial bump mixes incompatible internals |
| `production` | everything else that ships, patch + minor | Low volume; splitting it only adds another PR to the cascade |
| `dev-patch` / `dev-minor` | dev dependencies, split by update type | The dev side carries the volume (~25/week); the boring patches should land even while a minor is red |

**A 7-day `cooldown`** replaces the human as the filter against bad releases, mirroring the `minimumReleaseAge` this repo used to set in `pnpm-workspace.yaml` (`@mittwald/*` exempt for the same reason). Plus explicit `ignore` entries for the two deliberate holds, so they stop coming back weekly: `recharts >=3.8.0` (breaks the `CartesianChart` public type surface) and `playwright >=1.62.0` (WebKit 2336 stretches the LightBox image).

**`dependabot-auto-merge.yml`** waits for the visual suite, re-verifies the PR, and comments `@dependabot squash and merge`. Dependabot merges once the required `main` check is green.

**The visual suite now runs on every Dependabot PR** instead of on the `run-visual-tests` label. A `playwright`, `vitest` or `react-aria` bump moves snapshots — and a workflow cannot use the label route anyway: events created with `GITHUB_TOKEN` do not start new workflow runs. Its two writing steps are narrowed to the labelled path, because on a Dependabot PR the read-only token would 403 and fail the job — blocking the very merge it gates.

## ⚠️ This is inert until a maintainer changes the ruleset

The `main` ruleset requires a code owner approval, so nothing auto-merges until Dependabot has a bypass. A bypass list covers **every rule in its ruleset**, so ruleset `467449` has to be split in two — otherwise Dependabot could also skip CI:

| Ruleset | Rules | Dependabot bypass |
| --- | --- | --- |
| A | `pull_request` | yes |
| B | `required_status_checks`, `deletion`, `creation`, `non_fast_forward` | **no** |

Merging this PR without that split is safe — it just does nothing.

## Security trade-off, stated plainly

Trading the code owner approval for automation is deliberate. An approval on *"bump framer-motion 12.42.2 → 12.43.0"* does not detect a compromised release; nobody reads the dependency diff. The gates that actually can are mechanical: the cooldown, majors staying ignored, full CI, the visual suite, and a signature check. Given that merges to `main` publish on the standing prerelease line, the cooldown is the gate that matters most here.

The bypass opens exactly one hole: anyone with write access can push a commit onto an open `dependabot/*` branch, and `require_last_push_approval` no longer stops it. The workflow closes it — it merges only when **every** commit in the PR is authored by `dependabot[bot]` *and* carries a verified signature. It also refuses a stale green (a rebase landing mid-run), a major, and a command already issued for the same head commit.

## Known limits

- **`workflow_run` only fires from the default branch**, so the auto-merge workflow cannot be exercised from this PR. First real verification is the next Dependabot batch.
- **Whether Dependabot accepts merge commands from `github-actions[bot]` is undocumented.** GitHub's docs suggest `gh pr merge --auto` instead — but that merges as `github-actions[bot]`, which would then need the bypass, widening it to every workflow holding a write token. If Dependabot ignores the comment, the fix is a dedicated App token for the *comment*, not the broader bypass. There's a code comment saying so.
- **Human PRs will now show a skipped "Run Visual Regression Tests" run.** The author cannot be filtered at the `on:` level. Removing that cosmetic noise would mean refactoring the visual workflow into a `workflow_call` reusable plus two thin callers — deliberately not done here, because it would put churn on the label flow the team uses daily.
- **`CONTRIBUTE.md` ground rule 5 and `AGENTS.md` still claim pnpm enforces `minimumReleaseAge`.** It was deleted from `pnpm-workspace.yaml` in `de72c8037` (#2827, "prepare for typescript 7"), apparently by accident. Left untouched here because the fix depends on a decision — restore it or drop the claim — and is being filed as its own issue.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,22 +4,79 @@ version: 2
 # breaking changes (build/runtime/API) that warrant a considered upgrade rather
 # than an auto-PR. To take a specific major, bump it manually (or temporarily
 # remove/relax the matching ignore).
+#
+# Everything proposed here is auto-merged by
+# `.github/workflows/dependabot-auto-merge.yml` once CI and the visual
+# regression suite are green — see CONTRIBUTE.md § Dependency updates. Two
+# consequences for this file:
+#
+# 1. NOTHING IS UNGROUPED. Every npm dependency falls into one of the groups
+#    below, so a week produces at most four lockfile PRs. Each merge invalidates
+#    `pnpm-lock.yaml` in every other open PR, which makes Dependabot rebase it
+#    and CI run again — with ~10 open PRs that cascade is quadratic, and it is
+#    why the old `dev-dependencies` PR grew to 29 updates without ever landing.
+# 2. A `cooldown` replaces the human as the filter against bad releases. It
+#    mirrors the `minimumReleaseAge` this repo used to set in
+#    `pnpm-workspace.yaml`, `@mittwald/*` exempt for the same reason: our own
+#    packages are published and consumed within the hour.
 updates:
   - package-ecosystem: npm
     directory: "/"
     schedule:
       interval: weekly
-    open-pull-requests-limit: 10
+    open-pull-requests-limit: 5
+    cooldown:
+      default-days: 7
+      exclude:
+        - "@mittwald/*"
+    # First match wins, so the risk cluster is carved out before the catch-all.
     groups:
-      dev-dependencies:
+      # The accessibility foundation moves as one unit and breaks as one unit —
+      # a partial bump across these packages mixes incompatible internals.
+      react-aria:
+        patterns:
+          - "react-aria"
+          - "react-aria-components"
+          - "@react-aria/*"
+          - "@react-stately/*"
+          - "@react-types/*"
+          - "@internationalized/*"
+      # Everything that ships to consumers. Patch and minor together: with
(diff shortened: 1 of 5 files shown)


maaaathis (approved):
glhf

Issue #2887, opened 2026-08-25
Publish holes: versioned-but-unpublished releases break exact peer pins


mfal:
Our npm history has holes: versions that were bumped, changelogged and tagged but never published. Because `workspace:*` peer dependencies become **exact** pins at publish time, a hole is not cosmetic — it makes a published package demand a version that does not exist.

Reported as a side observation by an mStudio extension developer: `@mittwald/flow-remote-core@0.2.0-alpha.788` peer-requires `@mittwald/flow-react-components@0.2.0-alpha.787`, which was never published, so `pnpm install` warns and nothing can satisfy it. They worked around it by moving off that line. The cause was fixed in the July pipeline hardening — this issue is about what survived it, and about the guard that is still missing.

## What is actually in the registry

For released versions `>=0.2.0-alpha.600`, counting only versions that have a CHANGELOG heading in the package (i.e. lerna versioned them) but no artifact on npm:

| Package | versioned, never published |
| --- | --- |
| `@mittwald/flow-react-components` | 56 |
| `@mittwald/flow-remote-core` | 56 |
| `@mittwald/ext-bridge` | 50 |
| `@mittwald/flow-icons` | 52 |

Git tags exist for all of them (1048 `0.2.0-alpha.*` tags), so these are publish failures, not deliberate skips. `flow-icons` has a further 121 gaps with no changelog heading — those were never versioned, which is normal for a package that rarely changes.

Two shapes:

**A whole-release outage.** `alpha.898` (2026-07-09) through `alpha.961` (2026-07-28): roughly three weeks in which main kept versioning and tagging while nothing reached npm. Matches the workflow history exactly — `ci(publish): split versioning from publishing and add retry-safe from-package recovery` (#2658), `run lerna publish sequentially to stop sigstore flakes from orphaning packages` (#2714), `survive Rekor 409 and stop the version ratchet` (#2716, #2718).

**A partial release.** `alpha.787` published 5 of 12 packages:

```
published: ext-bridge, flow-design-tokens, flow-remote-core,
           mstudio-ext-react-components, react-tunnel
missing:   flow-react-components, flow-remote-elements,
           flow-remote-react-components, flow-remote-react-renderer,
           flow-icons, flow-icons-pro, flow-stylesheet
```

## Why a hole becomes a broken install

`peerDependencies` are declared as `workspace:*`, which pnpm replaces with the **exact** version at pack time:

```json
"peerDependencies": {
[...]

PR #2933, opened 2026-08-26
ci: skip the release for docs- and CI-only merges

### LINKED ISSUE #2931: Docs- and CI-only merges publish a full release of all 12 packages
Every push to `main` publishes all 12 packages, whether or not anything publishable changed. A third of our releases carry no package change at all.

## Evidence

Last 120 non-release commits on `main`:

- **40 (33 %)** touched only `.github/**`, `apps/**`, `docs/**` or root tooling files — 33 docs-only, 7 CI-only. Every single one produced a `chore(release):` bump and a full publish.
- 99 `chore(release):` bumps in the last 30 days × 12 published packages ≈ **1 200 npm versions/month**, roughly 400 of them for changes that ship no code.

Straight off `main`:

| Merge | What it touched |
| --- | --- |
| `chore(ci): purge CODEOWNERS def in favour of GitHub Teams` (#2906) | `.github/**` → `0.2.0-alpha.1055` |
| `docs: fix heading spacing` (#2870) | `apps/docs/**` → `0.2.0-alpha.1044` |
| `chore(docs): correct lint git hook to pre-push` (#2873) | one Markdown file → `0.2.0-alpha.1049` |
| `docs(sprach-guide): drop the "User" glossary entry` (#2850) | one Markdown file → `0.2.0-alpha.1032` |

## Why it happens

- `publish.yml` triggers on `push: main` with **no path filter**. It gates only on `vars.AUTO_PUBLISH_ENABLED` and the `chore(release):` self-push guard.
- The `Version` step runs `lerna version --force-publish`. `--force-publish` bypasses lerna's changed-package detection **by design** — fixed versioning means all 12 packages share one version, and #2887 traced the exact-peer-pin breakage to packages being skipped. So conventional-commits' "`docs`/`ci`/`chore` does not bump" rule never gets a chance to apply.

## Why this gets worse after the 1.0.0 cut

While `main` sits on `0.2.0-alpha.N`, the churn costs npm versions and reviewer noise. After the cut, the same path publishes a **stable `1.0.x` under `latest`** — a docs typo becomes a patch release every consumer picks up. Same on the `next` line: `forward-merge.yml` fires on the same push, and `publish-next.yml` re-derives and publishes a `next` prerelease.

## What to check

1. Confirm the numbers and map **every** trigger path that ends in a publish: `publish.yml`, and `forward-merge.yml` → `publish-next.yml`.
2. Pick the "is this publishable?" signal. None is obviously right:
   - **Path filter on the trigger** (`paths-ignore`) — cheapest, but globs are coarse. `packages/*/src/**/*.stories.tsx`, `**/AGENTS.md` and `**/*.test.ts` sit inside publishable packages and ship nothing. A skipped run is also invisible in the Actions list.
   - **Guard step inside `publish.yml`** that diffs the push range and exits early — expressive, one decision point, can `::notice::` why it skipped. Costs a runner start.
   - **`nx affected` against the publishable projects** — reuses the dependency graph the repo already trusts instead of a hand-maintained glob list.
   - **Commit type/scope** (`docs:`, `ci:`, `chore:`) — semantically cleanest, but unreliable here: `feat(docs): add a github link to the header` (#2855) and `fix(docs): make anchor links scroll to their heading` (#2881) are docs-only under `feat`/`fix`.
   - Dropping `--force-publish` is **not** a candidate on its own — that is exactly what produces publish holes (#2887).
3. Verify a skipped publish leaves no residue: no tag, no GitHub Release, no `chore(release):` commit. The next real release's changelog range must simply cover the skipped commits.

## Acceptance criteria

- A merge touching only `apps/docs/**`, `docs/**`, `.github/**` or root tooling produces **no** npm publish, no tag, no GitHub Release, no `chore(release):` commit on `main`.
- A merge touching any publishable package source publishes exactly as today.
- A **mixed** merge (docs + package source) publishes.
- `workflow_dispatch` still forces a publish regardless of paths — the manual escape hatch stays.
- The decision is visible in the run: a skip says which rule fired.
- The `next` line behaves symmetrically, or the difference is documented.
- The rule is written down next to the release model (ADR 0005 / #2711), so the next person does not rediscover it.

## Adjacent, not in scope

`deploy-main.yml` rebuilds and redeploys both Docker images (docs **and** storybook) on every push to `main` — including CI-only ones. Same root cause, much smaller blast radius. Separate issue if we want it.

Related: #2711, #2816, #2887, ADR 0005.

mfal:
## What & why

Closes #2931.

Every push to a release line published all 12 packages, whether or not anything publishable changed. Over the last **200** non-release commits on `main`, **76 (38 %)** touched only `apps/**`, `docs/**`, `.github/**` or root tooling — each produced a `chore(release):` bump, a tag, a GitHub Release and 12 npm versions that ship no code. After the 1.0.0 cut the same path would publish a stable `1.0.x` under `latest` for a docs typo.

Two causes, both deliberate on their own: `publish.yml` had no path filter, and `lerna version --force-publish` bypasses lerna's changed-package detection (fixed versioning — #2887 traced the exact-peer-pin breakage to packages being skipped). So conventional-commits' "`docs`/`ci`/`chore` does not bump" rule never got a chance to apply.

A `decide` job now classifies the pushed file set and gates the publish job through `needs`. **A skipped push produces no npm publish, no version bump commit, no tag and no GitHub Release.**

## What a reviewer should know

**Why a job and not `paths-ignore`.** The run shows `decide` green and `publish` skipped, with a notice naming the rule that fired. `paths-ignore` would suppress the run and explain nothing — and its globs cannot express that Markdown *inside* `packages/` still ships: `@mittwald/flow-react-components` publishes `["*.md", "dist"]`.

**The rule is a denylist, on purpose.** A forgotten docs path costs one needless version; a forgotten source path would silently swallow a real release. So unknown paths count as publishable and `packages/**` counts wholesale. Dependency bumps (`pnpm-lock.yaml`) therefore still publish — they change real deps.

**`workflow_dispatch` always publishes** — the escape hatch when a docs-only change has to go out as a release anyway.

**Everything unreadable publishes.** No usable base sha, a failed compare call, or a response that is not a list of paths → publish. `gh api` prints its error body to *stdout* on a non-2xx, so the exit status decides whether the list is usable, and the classifier rejects a JSON blob outright instead of counting it as a filename.

**`publish-next.yml` gets the same gate**, so a forward-merged docs fix skips on `next` exactly as it skipped on `main`.

**This PR demonstrates itself:** it is CI-only, so merging it triggers no release.

## Verification

The shipped step script was extracted from the YAML (not re-typed) and run against real commits and the real compare API:

| Case | Result |
| --- | --- |
| `docs: fix heading spacing` (#2870) | `publish=false` |
| `fix(Link): stick link icon…` (#2869) | `publish=true` (7/7 relevant) |
| `chore(ci): purge CODEOWNERS def` (#2906) | `publish=false` |
| unreachable base sha | `publish=true` (fail-safe) |
| branch creation / force push | `publish=true` (fail-safe) |
| manual dispatch | `publish=true` |

Replayed over 200 non-release commits on `main`: **76 would be skipped, 0 false skips** — no commit touching `packages/**` is ever skipped.

`test.yml` now runs `node --test .github/scripts/*.test.mjs`, so a broken classifier fails the PR rather than the next release. The guard scripts sit outside the nx graph, so `pnpm affected:test` never saw them.

## Checklist

- [x] `pnpm lint` is clean and `pnpm affected:test` passes (browser tests if
      behavior changed)
- [x] **Generated code is committed** (`git diff` is empty after the relevant
      `build:*` targets)
- [x] User-facing strings added to **both** `de-DE` and `en-US` locale files
- [x] Docs updated if a public API changed; intentional visual changes get
      updated snapshots / the `update-screenshots` label

<sub>No package code changes, so no locale, snapshot or generated-code work applies; `pnpm lint` is clean and the 24 guard tests pass.</sub>

### DIFF
--- a/.github/scripts/release-relevance-guard.mjs
+++ b/.github/scripts/release-relevance-guard.mjs
@@ -1,0 +1,64 @@
+#!/usr/bin/env node
+// @ts-check
+/**
+ * Release-relevance guard — IO shell around release-relevance-lib.mjs.
+ *
+ * Reads the changed-file list on stdin (one repository-relative path per line)
+ * and writes `publish=true|false` to $GITHUB_OUTPUT.
+ *
+ * It never fails the run. The decision is a routing choice, not a policy
+ * violation: an empty or unreadable input yields `publish=true`, which is the
+ * behaviour before #2931.
+ */
+import { appendFileSync, readFileSync } from "node:fs";
+import { classifyChangedFiles } from "./release-relevance-lib.mjs";
+
+/** Read all of stdin; an unreadable stdin is an empty list (fail-safe). */
+function readStdin() {
+  try {
+    return readFileSync(0, "utf8");
+  } catch {
+    return "";
+  }
+}
+
+const files = readStdin()
+  .split("\n")
+  .map((line) => line.trim())
+  .filter((line) => line !== "");
+
+const { publish, reason, relevant, total } = classifyChangedFiles(files);
+
+// Cap the log: a large merge lists hundreds of files and the first few already
+// explain the decision.
+const SAMPLE = 20;
+
+if (publish) {
+  // `relevant` is empty on the fail-safe paths (unreadable file list), where
+  // "publishable change" would overstate what we actually know.
+  const headline = relevant.length > 0 ? "Publishable change" : "Publishing";
+  console.log(`::notice::${headline} — ${reason}.`);
+  if (relevant.length > 0) {
+    console.log(`Files that reach a published package (${relevant.length}):`);
+    for (const path of relevant.slice(0, SAMPLE)) console.log(`  ${path}`);
+    if (relevant.length > SAMPLE) {
+      console.log(`  … and ${relevant.length - SAMPLE} more`);
+    }
+  }
+} else {
+  console.log(
+    `::notice::No publishable change — ${reason}. ` +
+      "This push produces no npm publish, no version bump commit, no tag and " +
+      "no GitHub Release. Dispatch this workflow manually to publish anyway.",
+  );
+  console.log(`Changed files (${total}):`);
+  for (const path of files.slice(0, SAMPLE)) console.log(`  ${path}`);
+  if (files.length > SAMPLE) {
+    console.log(`  … and ${files.length - SAMPLE} more`);
+  }
+}
+
+const out = process.env.GITHUB_OUTPUT;
+const line = `publish=${publish}\n`;
+if (out) appendFileSync(out, line);
+else process.stdout.write(line);
--- a/.github/scripts/release-relevance-lib.mjs
+++ b/.github/scripts/release-relevance-lib.mjs
@@ -1,0 +1,155 @@
+// @ts-check
+/**
+ * Release-relevance classification — pure functions, no git / no IO.
+ *
+ * Answers one question: can this set of changed files reach a PUBLISHED
+ * package's tarball? Docs- and CI-only merges cannot, and must not produce a
+ * release — no npm publish, no `chore(release):` bump, no tag, no GitHub
+ * Release (#2931).
+ *
+ * The rule is a DENYLIST, and that direction is deliberate. Every published
+ * package ships `dist` (plus `*.md` for flow-react-components), so "what ends
+ * up in a tarball" spans nearly all of `packages/**` — source, locales, SCSS,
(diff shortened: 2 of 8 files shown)


maaaathis (approved):
glhf

Issue #2934, opened 2026-08-26
Docs: deprecated components in the ungrouped list and their successor links


mfal:
Deprecated components are handled inconsistently in the docs. Three small UX gaps, all around the same theme: make a deprecated component recognizable as such and point the reader at its successor.

### 1. Sink deprecated components to the bottom in the ungrouped navigation

The grouped component navigation already ranks deprecated entries last — `deprecatedRank` in `sortEntries` (`apps/docs/src/app/_components/layout/MainNavigation/MainNavigation.tsx:78`).

The alphabetical (ungrouped) view does not: it sorts `collectMdxFiles(componentEntries)` by label only (`MainNavigation.tsx:165`), so `Align` and `SegmentedControl` sit among the active components. Switching the grouping mode silently changes the ranking rule.

Apply the same `deprecatedRank` tiebreaker there, so deprecated components land at the bottom of the flat list in both views. The `Deprecated` badge already renders in both.

### 2. Status badges on the overview cards, deprecated ones last

`ComponentsOverview` feeds the pre-sorted list into both views (`apps/docs/src/app/04-components/_components/ComponentsOverview.tsx`) and never ranks deprecated last — neither grouped nor alphabetical. `ComponentCard` renders name, description and wireframe, but no status at all, so a deprecated component is indistinguishable from an active one.

- Render `ComponentStatusBadge` on the card (`apps/docs/src/app/04-components/_components/ComponentCard.tsx`) — the same component the navigation already uses, so Beta/Deprecated/Neu stay consistent across nav and overview.
- Rank deprecated cards last in both views, same tiebreaker as item 1.

### 3. Link the replacement component from the deprecation notice

`ComponentStatusCallout` renders the page's `deprecationNotice` frontmatter as plain text (`apps/docs/src/lib/componentStatus/ComponentStatusCallout.tsx:42`). Every notice we have names a successor, and none of them is clickable:

- `SegmentedControl` → "Nutze Tabs … und eine RadioGroup" (`src/content/04-components/form-controls/segmented-control/index.mdx`)
- `Align` → "nutze stattdessen Combine" (`src/content/04-components/structure/align/index.mdx`)

The reader has to go back to the navigation and search for the name. The successor is the single most useful link on a deprecated page.

[...]

PR #2940, opened 2026-08-26
docs: link the successors of a deprecated component

mfal:
## What & why

Part of #2934. The issue lists three gaps around deprecated components; **this
PR now covers only the third**, because @Lisa18289 built the other two in #2937
and hers is the better version — see below.

**The successor was not clickable.** The deprecation callout rendered its notice
as plain prose, so the one thing a reader needs next — the replacement component
— meant a trip back through the navigation. A new `replacedBy` frontmatter field
lists the successors by display name; `ComponentContent` resolves them against
the component pages and the callout renders them as links.

```
replacedBy: [Tabs, RadioGroup]
```

Aligns notice lost its now-redundant trailing "– nutze stattdessen Combine",
since the successor is a link now.

## What was dropped, and why

The first version also badged the overview tiles and ranked deprecated
components last in both views. #2937 does both, further:

- Its badge sits in the tile's **top right corner** — the place the design gives
  status badges — where this one sat inline next to the title. It also handles a
  stacking problem this version never hit: an absolutely positioned badge needs
  the tile to `isolate`, or it paints over the sticky header while scrolling.
- Its ordering **consolidates navigation and overview onto one `compareEntries`**
  (deprecated, then numeric path prefix, then label) and fixes the
  integration-group split on the way. This version added a deprecated-last
  tiebreaker to each comparator separately — but two comparators that must agree
  are exactly what let the order drift apart, so one comparator is the real fix.

Both were removed in 144c36ad1 rather than force-pushed away, so the review
thread on the original stays readable. #2934 is fully addressed once both merge.

## Notes for reviewers

- **`replacedBy` resolution throws on an unknown name.** A typo would otherwise
  silently drop the most useful link on the page, and the docs build renders
  every component page, so a bad name fails CI instead of shipping.
- Reading the sibling pages moved `ComponentContent` from
  `MdxFileFactory.fromParams` to `fromDir` + a slug match — the successors have
  to resolve against something, and the component pages are that something.
- **Base is `main`, type is `docs:`** — documentation-site only, nothing
  published to npm.
- No overlap with #2937 left: this PR touches the callout, `replacedBy.ts`,
  `MdxFile` meta and two `.mdx` frontmatters. Merge order does not matter.

## Checklist

- [x] `pnpm lint` is clean and `pnpm nx build docs` passes — the docs app has no
      test target, the build is the gate, and it type-checks and renders every
      component page
- [x] **Generated code is committed** — no generators involved
- [x] User-facing strings: N/A — docs-site prose (German, like the rest of
      `apps/docs`), not component locale files
- [x] No component API and no component rendering changed, so no visual
      snapshots are affected

### DIFF
--- a/apps/docs/src/app/04-components/_components/ComponentCard.tsx
+++ b/apps/docs/src/app/04-components/_components/ComponentCard.tsx
@@ -3,7 +3,6 @@ import type { FC } from "react";
 import { Heading, Text } from "@mittwald/flow-react-components";
 import type { ComponentLink } from "@/app/04-components/_components/ComponentsList";
 import { getWireframe } from "@/app/04-components/_components/wireframe/registry";
-import { ComponentStatusBadge } from "@/lib/componentStatus";
 import styles from "./ComponentCard.module.scss";
 
 interface Props {
@@ -20,14 +19,9 @@ export const ComponentCard: FC<Props> = (props) => {
         <Wireframe />
       </div>
       <div className={styles.content}>
-        <div className={styles.header}>
-          <Heading level={3} elementType="p" className={styles.title}>
-            {component.name}
-          </Heading>
-          {component.component && (
-            <ComponentStatusBadge name={component.component} />
-          )}
-        </div>
+        <Heading level={3} elementType="p" className={styles.title}>
+          {component.name}
+        </Heading>
         {component.description && (
           <Text className={styles.description}>{component.description}</Text>
         )}
--- a/apps/docs/src/app/04-components/_components/ComponentCard.tsx
+++ b/apps/docs/src/app/04-components/_components/ComponentCard.tsx
@@ -3,6 +3,7 @@ import type { FC } from "react";
 import { Heading, Text } from "@mittwald/flow-react-components";
 import type { ComponentLink } from "@/app/04-components/_components/ComponentsList";
 import { getWireframe } from "@/app/04-components/_components/wireframe/registry";
+import { ComponentStatusBadge } from "@/lib/componentStatus";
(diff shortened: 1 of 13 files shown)


apps/docs/src/app/04-components/_components/ComponentCard.tsx
```diff
@@ -19,9 +20,14 @@ export const ComponentCard: FC<Props> = (props) => {
...
       <div className={styles.content}>
-        <Heading level={3} elementType="p" className={styles.title}>
-          {component.name}
-        </Heading>
+        <div className={styles.header}>
+          <Heading level={3} elementType="p" className={styles.title}>
+            {component.name}
+          </Heading>
+          {component.component && (
+            <ComponentStatusBadge name={component.component} />
```
Lisa18289: das hab ich auch grad in einem anderen mr eingebaut 😅 hast du was dagegen wenn wir da meine Variante nehmen? da liegen die Badges an der dafür vorgesehenen Stelle ganz oben rechts in der Kachel

mfal: Sehr gerne — deine Variante ist besser, ich habe das Badge und die Sortierung hier komplett rausgenommen (144c36ad1). Nicht nur die Badge-Position: dein `compareEntries` führt Navigation und Overview auf einen Comparator zusammen, statt in beiden einzeln einen Deprecated-Tiebreaker zu ergänzen. Übrig bleibt hier nur `replacedBy` + die verlinkten Nachfolger im Callout, das berührt #2937 nicht.

Lisa18289 (approved):

PR #2945, opened 2026-08-26
fix(CodeBlock): balance the padding and unify the editor line colors

Lisa18289:
Two visual fixes in the code block.

**Padding.** With the `code` prop the block renders a `CodeEditor`, which is a form control and took its padding from the form control tokens — 7px block, 16px inline. `--code-block--padding` only reached the `children` variant, so both variants of the component disagreed. Both axes now use the component's own token.

**Line colors.** The gutter border (`#ddd`, CodeMirror) and the indentation markers (`#F0F1F2`, `@replit/codemirror-indentation-markers`) were hard coded light theme values — the marker library picks by CodeMirror's dark theme flag, which `githubLightInit` never sets, so both stayed bright on the dark surface. Two new tokens replace them:

- `code-editor.line-color` → `separator.color`, the token the code block draws its border with, so the field border and the gutter border share one color.
- `code-editor.indentation-marker-color` → `neutral.color.400`, one step closer to the surface, so the guides inside the field stay below the lines that bound it.

Both follow the theme, and the active indentation marker loses its own color.


packages/design-tokens/src/form/code-editor.yml
```diff
@@ -5,3 +5,7 @@ code-editor:
     value: "FiraCode, monospace"
   gutter-element-min-width:
     value: "{size-rem.l}"
+  line-color:
```
Lisa18289: nenn das gutter-separator-color

Lisa18289: Renamed to `code-editor.gutter-separator-color` in 5080dc081 — it only draws the gutter separator now that the indentation markers have their own token. Value unchanged.

maaaathis (approved):

PR #2954, opened 2026-08-27
docs(ColumnLayout): explain breakpoints through a resizable example

Lisa18289:
The ColumnLayout page explained its breakpoint behaviour in prose, split across "Responsive Layout" and "Benutzerdefinierte Werte", and both shared a single example. A reader could not see the behaviour, because a live example renders at whatever width the docs content column happens to have.

**Docs infrastructure:** a new `resizable` prop on `LiveCodeEditor` frames the preview and puts a handle on its right edge. Dragging it sets the container width; a caption names the resulting breakpoint. The handle is keyboard operable (arrow keys, Home, End).

**Page:** the prose is split into a reference section and a demonstration.

- `# Breakpoints` — thresholds, default values and the inheritance rule, no example
- `# Spalten` — what `s`, `m` and `l` set, through a resizable example
- `# Ausgeblendete Spalten` — reduced to what `null` does, real content instead of lorem ipsum
- `# Abstände` — gained the missing `xl`, lost the pixel values that no other component page states either

`custom-values.tsx` is renamed to `columns.tsx` to match the section that uses it.

Two findings left out of this PR: the properties table renders `s`/`m`/`l` as `(number)[]` instead of `(number | null)[]`, and `gap` as the opaque type name `GapSize` rather than its allowed values. Both come from `build:docs-properties`.


apps/docs/src/content/04-components/structure/column-layout/index.mdx
```diff
@@ -16,67 +16,50 @@ description:
...
-`m`, anstatt auf den Standardwert von `[1, 1, 1]` zurückzufallen. Da für den
-Breakpoint `s` kein benutzerdefinierter Wert angegeben wurde, bleibt er beim
-Standardwert von `[1]`.
-
-<LiveCodeEditor example="custom-values" editorCollapsed />
+<LiveCodeEditor example="columns" resizable editorCollapsed />
 
 ---
 
 # Ausgeblendete Spalten
```
Lisa18289: das soll eine sub heading von spalten sein

Lisa18289: Done in 55765a8 — `Ausgeblendete Spalten` is now a `##` under `# Spalten`, and the `---` between them is gone (per the README, no rule between a `#` and its own `##`). It still shows up in the anchor navigation.

mfal (approved):

Issue #2774, opened 2026-08-04
RFC: Public export surface — components in the root barrel, everything else addressed


mfal:
## Proposal

Restructure the public export surface of `@mittwald/flow-react-components`:

- **Root barrel (`.`)** — components and their props interfaces only.
- **`/<ComponentName>`** — extras that belong to exactly one component (`typedList`, `typedCartesianChart`, …).
- **Thematic sub-barrels** — cross-cutting, non-component API: `/theming`, `/hooks`, `/props-context`. Not a single catch-all `/lib`.
- **`/internal`** — infrastructure that should never have been exported.

Today `.` is a grab bag. Next to ~110 components it also exports theming (11 symbols), 6 hooks, propsContext (4), the overlay controller, and `flags` / `useControlledRemoteValueProps` / `eventHandlerContext`. Those last three have no place in a consumer entrypoint. This proposal replaces the grab bag with a rule.

## Hard constraint: no breaking changes in 1.0.0

Product management agreed: **1.0.0 ships no breaking changes.** A major version number is not a licence to break things. And a codemod is not an excuse either — it lowers the cost of a migration we force on consumers, but it doesn't make that migration free, and it does nothing for consumers who never run it.

So the plan is simple: **add addresses, remove nothing.** Every symbol keeps its root-barrel export. New subpaths are additive. `@deprecated` markers point at the preferred path. Removals wait for 2.0.0.

This rules out the most valuable part of the RFC — see [What the constraint costs us](#what-the-constraint-costs-us).

## Measured baseline

| | |
| --- | --- |
| Symbols in the root barrel (`src/index/default.ts`) | **511** (132 of them icons → **379** without icons) |
| Component modules re-exported by `src/components/public.ts` | **110** |
| …of which have non-component, non-`*Props` exports | **9** (26 symbols) |
| `exports` map entries today | 9 |
| Import profile in `apps/docs` | 603× root vs. 42× subpath |
| Import profile in mStudio (`mstudio-next`, 3738 files) | 1979× root, 413× `/react-hook-form`, 15× `/password-tools` |

The 9 modules with extras: `Action`, `List`, `CartesianChart`, `NotificationProvider`, `TranslationProvider`, `Icon`, `DeprecationWarningProvider`, `CountryOptions`, `RangeCalendar`.

## Consumer usage — measured against mStudio

[...]

PR #2960, opened 2026-08-27
fix(docs): hide deprecated values from the properties table

Lisa18289:
A prop can keep a value working while no longer wanting anyone to use it — `Button`'s `color="accent"`, renamed to `success` long ago, was still listed in the properties table as a normal choice.

`@deprecatedValues accent` on the prop's JSDoc drops it from the table. The mechanism is generic: `react-docgen-typescript` already carries the tag through to `doc-properties.json`, so any prop can list one or more deprecated values, and the same top-level union filter that removes `undefined`/`null` (from #2961) removes them.

```
color   … | success | danger | primary (default) | secondary | accent
    →   … | success | danger | primary (default) | secondary
```

Nothing about the prop itself changes: `accent` stays in `ButtonProps["color"]`, still renders as `success`, and still warns via `useWarnDeprecation`. The deprecation window stays open — the table just stops advertising it.

Based on #2961, whose union splitter this reuses.

<sub>This PR previously removed `accent` from the type instead; that version is dropped, see the discussion below.</sub>

### DIFF
--- a/apps/docs/src/lib/PropertiesTables/lib/loadProperties.ts
+++ b/apps/docs/src/lib/PropertiesTables/lib/loadProperties.ts
@@ -1,23 +1,34 @@
 import docGenFile from "@mittwald/flow-react-components/doc-properties";
 import type { ComponentDoc } from "react-docgen-typescript";
 import type { Properties, Property } from "../types";
-import { splitUnion } from "./unionType";
+import { splitUnion, unquote } from "./unionType";
 
 const eventRegex = /^on[A-Z]+.*/;
 const a11yRegex = /^aria-.+/;
 const nullishMembers = ["undefined", "null"];
+/** `@deprecatedValues accent, plain` — but not the prop-level `@deprecated`. */
+const deprecatedRegex = /@deprecated(?!\w)/;
+const deprecatedValuesRegex = /^[ \t]*@deprecatedValues[ \t]+(.*)$/m;
 
 /*
- * `undefined` and `null` as top-level union members only restate the "Required"
- * badge. Nested ones are part of the type and have to survive: ColumnLayout's
- * `(number | null)[]` documents that `null` hides a column.
+ * Drops union members the table must not offer: `undefined`/`null`, which only
+ * restate the "Required" badge, and the values a prop marks `@deprecatedValues`
+ * — those still work at runtime, but nothing should send a reader towards them.
+ * Only top-level members go; nested ones are part of the type and have to
+ * survive, e.g. ColumnLayout's `(number | null)[]`, where `null` hides a column.
  */
-const stripTopLevelNullish = (type: string): string => {
+const hideMembers = (type: string, hidden: string[]): string => {
   const members = splitUnion(type).filter(
-    (member) => !nullishMembers.includes(member),
+    (member) => !hidden.includes(unquote(member)),
   );
   return members.length > 0 ? members.join(" | ") : type;
 };
+
+const parseDeprecatedValues = (description: string): string[] =>
+  (deprecatedValuesRegex.exec(description)?.[1] ?? "")
+    .split(",")
+    .map((value) => unquote(value.trim()))
+    .filter(Boolean);
 
 /** An `@default: x` JSDoc tag keeps its colon in the generated metadata. */
 const normalizeDefaultValue = (value: unknown): string | null => {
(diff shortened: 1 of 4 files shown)


mfal:
Two things, both about the shape of this rather than the code.

## 1. One value on one component — where is the general mechanism?

Value-level deprecation is not a Button special case:

- `AccentBox.color` accepts `"neutral"`, `"gradient"`, `"green"` through the same
  kind of cast (`AccentBox.tsx:74`, `:86`) — and warns for none of them.
- `CartesianChart.emptyView` deprecates a *value shape* (non-element) and warns.
- Every future rename of a variant, color or status adds one more.

What this PR codifies in AGENTS.md is "write a cast at each comparison site".
That is exactly the part that does not scale, and this PR already shows it: three
sites (`Button`, `Chat`, `getActionGroupSlot`), three hand-written casts that
already diverge (`as string` vs. `as string | undefined` — `Chat`'s `as string`
lies, `color` is optional). They also go stale silently: when the runtime
handling is finally removed, nothing tells you the cast is dead code, because a
cast to `string` compares fine against any literal forever.

There is a pattern for this in the repo already —
`ComponentDefaultsProvider/deprecatedFlags.ts`: a small registry of rows plus one
hook that warns with the standard message. The same shape fits here:

```
{ component: "Button", prop: "color", value: "accent", replacement: "success" }
```

- `useDeprecatedPropValue(...)` returns the normalised value and emits the
  standard message — wording enforced by construction, not by copy-paste.
- `isDeprecatedValue("Button", "color", x)` replaces the ad-hoc casts in `Chat`
  and `getActionGroupSlot`.
- At removal time, one row disappears instead of a cast hunt.
- The same row can feed `MIGRATION.md` and the codemod list.

And the half of the problem that actually annoys users — the properties table
offering `accent` as a normal choice — needs no type change at all.
`doc-properties.json` comes from `react-docgen-typescript`, which already carries
prop-level JSDoc tags. A tag on the prop (`@deprecatedValues accent`) lets the
table drop or strike the value, generically, for every component.

If we don't want the general mechanism in this PR: fine. But then AGENTS.md
should not present the per-site cast as *the* pattern.

## 2. It is a TypeScript break, and "best effort" points the other way

ADR 0005 §4 says the type surface is not semver-protected. It does not say "break
it when convenient". Best effort means we *try* to keep types working; §4 exists
so a type break can ride along as fallout of a real change. Here the break is the
entire content of the change.

The result contradicts itself inside one PR: the runtime deliberately keeps
`accent` working ("keeps working until it is removed"), while the type says it is
removed *now*. A TS consumer gets a failing build from a **patch** release for a
value we simultaneously promise still works. Deprecation means there is a window;
a type error closes that window immediately, and precisely for the consumers who
care most (TS + CI). Removing the value is what the major line is for — then
type, runtime, docs and codemod land as one message (cf. #2774, 1.0.0 stays
additive).

On the payoff: I checked whether the value can stay in the type while completion
stops offering it. It cannot. `@deprecated` via a deprecated alias in a union is
not reflected in value completion (TS 6.0.3, `kindModifiers: ""`), while a
deprecated *property* is (`kindModifiers: "deprecated,optional"`). So the entire
gain of this break is editor completion — and the properties table, the more
visible half, is fixable without it.

One process detail if the break ships anyway: `refactor` is `hidden: true` in the
conventionalcommits preset (`DEFAULT_COMMIT_TYPES`). The commit still bumps a
patch, but it never appears in `CHANGELOG.md` or the release notes — so ADR 0005
§4's requirement (migration note in the commit body **and** in the release notes)
is not met as titled. `fix(Button): …` would put it in front of consumers.

My preference: drop the type change, fix the table generically (JSDoc tag +
generator), and keep `accent` typed until the removal. Keeping the type change
means at minimum a retitle, and ideally routing it through the shared mechanism
from point 1 instead of three casts.

Lisa18289:
Dropped the type change — you're right, and the self-contradiction you named is the clearest argument: the runtime deliberately keeps `accent` alive while the type removed it *now*, and a TS consumer would have gotten a failing build out of a patch release for a value we promise still works.

What is left is your point 2: `@deprecatedValues accent` on the prop's JSDoc, and `loadProperties` drops those members through the same top-level union filter that already removes `undefined`/`null` (#2961). Generic, one tag per prop, any number of values. `accent` stays in `ButtonProps["color"]`, still renders as `success`, still warns. Four files, no `MIGRATION.md` entry, no codemod pressure — the deprecation window stays open.

That also takes the three casts back out, so point 1's scaling problem is gone *from this PR* — nothing compares against an off-type value any more, and the `as string` / `as string | undefined` divergence you spotted goes with it.

The registry idea still stands for the **runtime** half, though, and your `AccentBox` observation is the real bug there: three deprecated color values, no warning for any of them. `CartesianChart.emptyView` is a third shape again. Shall I open a separate issue for `useDeprecatedPropValue` plus those gaps? It would pair well with the tag — same row could feed both the warning and the JSDoc.

Point 4 is handled by the retitle: `fix` has no `hidden` flag in the preset (only `refactor`, `docs`, `chore`, … do), so this lands under Bug Fixes in the changelog.

Two notes on the mechanics: #2961 merged while this was being reworked, so the PR now sits on `main` directly instead of stacking. And the branch name is stale — GitHub won't let me repoint the head branch of an open PR, and a fresh PR would drop this thread, so I left it.

mfal (approved):

PR #2957, opened 2026-08-27
docs: fix broken links and add missing Component cross-references

mfal:
## What & why

Two related gaps in the Styleguide content, both about links.

**Four internal links were dead** (404 on flow.mittwald.de):

| Page | Dead target | Now |
| --- | --- | --- |
| Forms pattern | `/04-components/form-controls/form-react-hook-form` | `/04-components/react-hook-form/form` |
| Forms pattern | `/04-components/form-controls/field-react-hook-form` | `/04-components/react-hook-form/field` |
| BigNumber | `/04-components/content/rating` | `/04-components/form-controls/rating` |
| TabNavigation | `/04-components/status/alert-icon` | link removed |

`AlertIcon` has no docs page at all — the Component exists and is used in several examples, but nothing documents it. Per [README](https://github.com/mittwald/flow/blob/main/apps/docs/README.md) ("Write the name in plain text … when no page exists for it") the name is now plain text. **If that page is supposed to exist, it needs its own issue** — this PR does not add it.

**32 places named a Component in prose without linking it** (50 links added), which the README explicitly asks for: *"Use inline links generously to connect related documentation. Component names in prose link to the Component's page."* The Boundaries page was the worst case — eleven Components mentioned, not one linked.

## What a reviewer should know

The candidate list was heavily polluted, so **every hit was read in context instead of replaced mechanically**. Of 113 raw matches, 57 were German words colliding with Component names and were deliberately left alone:

- **all 29 `Navigation`** — the activity ("die Navigation innerhalb des Seitenbaums"), never the Component
- **all 12 `Color`** — the color or token, and those sentences already link to Foundations › Color, which is the correct target
- **`Form`** — mostly "Form Control" (a category) or the German "in Form von"
- **`Accordion`** in [list/index.mdx](https://github.com/mittwald/flow/blob/main/apps/docs/src/content/04-components/structure/list/index.mdx) — that is List's own `accordion` property; a link would send readers to the unrelated Accordion Component

Where a Component is mentioned several times on a page, only the **first** occurrence is linked, per the README rule for repeated mentions.

Also removed a stray `[Section]` link in `02-informationskonzept` that rendered as `[Section][Headings]` with no separator (requested during review of the findings).

No new content, no new examples, no prose rewording beyond the linking itself.

## Verification

- `BROKEN internal links: 0` — every internal link target resolves to an existing route
[...]

PR #2971, opened 2026-08-27
docs(AlertIcon): add the missing Component page

mfal:
AlertIcon had no Styleguide page, while its siblings [Alert](https://flow.mittwald.de/04-components/status/alert), [AlertBadge](https://flow.mittwald.de/04-components/status/alert-badge) and [AlertText](https://flow.mittwald.de/04-components/status/alert-text) all have one. The gap cost a cross-reference: the AlertIcon link on the TabNavigation page pointed nowhere and was de-linked to inline code in #2957, when the docs link check surfaced it.

## What changed

- **New page** `apps/docs/src/content/04-components/status/alert-icon/index.mdx`, following the fixed spine — default example, `# Best Practices`, the shared visual axes (`# Status`, `# Sizes`), `# Kombiniere mit ...`, `# Properties`.
- **Four examples** (`default`, `status`, `sizes`, `tab-navigation`) with mittwald-domain content.
- **Link restored** in `tab-navigation/index.mdx:52` — an exact revert of the de-link from #2957.

## Content decisions

- **Best Practices** name the rules that exist only for this Component: the icon alone states neither cause nor remedy, so the state is explained where the icon points to; only the decision-relevant status is shown when several compete; and Alert, AlertBadge, AlertText and Notification already render their own status icon, so a second one is wrong inside them (verified in the component sources).
- **`## AlertIcon vs. AlertBadge`** is the one legitimate `<DoAndDont>` case per the content guidelines: the pair differs in whether the status is *named* — a real choice, not a visual one.
- **The automatic, localized `aria-label`** is documented as automatic behaviour in the `# Status` section, next to what it describes, rather than in a separate accessibility block.

## Verification

- Rendered against `pnpm nx dev docs`: the page returns 200, all four live examples render (the icons carry their automatic `aria-label`s), the properties table resolves `status`/`size`/`color`, and AlertIcon now appears in the sidebar navigation and on the component overview. The TabNavigation page emits `href="/04-components/status/alert-icon"`.
- Prettier clean (80-char prose, 60-char examples).

## Note for the reviewer

`pnpm nx test:links docs` does not exist on `main` yet — the link check is still unmerged work on another branch. I ran it here by copying `src/lib/links/` and `src/lib/mdx/anchors.ts` in temporarily and executing the test with `tsx`, then removed the copies (nothing of it is in this diff). Result: the AlertIcon link resolves. The only remaining break is the pre-existing missing leading slash on `04-components/status/notification-provider` in `02-informationskonzept/index.mdx:78` — already fixed on that same unmerged branch, so I left it alone here rather than duplicating the fix.

Optional follow-up, not done here: the Informationskonzept's „Status-Icon" section still links to the icon library rather than to the new AlertIcon page.

### DIFF
--- a/apps/docs/src/content/04-components/navigation/tab-navigation/index.mdx
+++ b/apps/docs/src/content/04-components/navigation/tab-navigation/index.mdx
@@ -49,8 +49,9 @@ dargestellt.
 
 # Mit Status-Icons
 
-Innerhalb der Links kann ein `AlertIcon` verwendet werden, um auf Zustände
-innerhalb eines Navigationsbereichs hinzuweisen.
+Innerhalb der Links kann ein [AlertIcon](/04-components/status/alert-icon)
+verwendet werden, um auf Zustände innerhalb eines Navigationsbereichs
+hinzuweisen.
 
 <LiveCodeEditor example="status" editorCollapsed stretch />
 
--- a/apps/docs/src/content/04-components/status/alert-icon/examples/default.tsx
+++ b/apps/docs/src/content/04-components/status/alert-icon/examples/default.tsx
@@ -1,0 +1,3 @@
+import { AlertIcon } from "@mittwald/flow-react-components";
+
+<AlertIcon status="warning" />;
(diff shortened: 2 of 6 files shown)


mfal:
Wird nicht umgesetzt: AlertIcon soll keine eigene Seite bekommen. Die Component wird überwiegend intern verwendet (Alert, AlertBadge, AlertText, Notification rendern sie selbst); die einzige öffentliche Stelle ist die TabNavigation, wo sie im Abschnitt "Mit Status-Icons" bereits erwähnt ist. Damit bleibt auch der Link dort bewusst entlinkt (siehe #2957).

Issue #3018, opened 2026-08-31
ImageCropper turns a failed image load into an uncaught rejection; its docs examples cannot load locally


mfal:
`/04-components/upload/image-cropper` is the only docs page that produces uncaught errors.

## The defect

When the image cannot be loaded, `ImageCropper` leaves the rejection unhandled:

```
Uncaught (in promise) Error: Failed to load image
    at image.onerror (…components_src_components_….js)
```

Four times on that page — one per example. There is no error state, no fallback, nothing the consuming app can react to. A cropper pointed at a URL that 404s, expires, or is blocked will do this in production too, so this is not a docs-only problem.

## What triggers it locally

The examples load the image from the deployed docs site:

```tsx
<ImageCropper image="https://flow.mittwald.de/assets/mittwald_logo_rgb.jpg" />
```

react-easy-crop reads the image into a canvas, which needs CORS. Same-origin on flow.mittwald.de, cross-origin from `localhost:3000`:

> Access to image at 'https://flow.mittwald.de/assets/mittwald_logo_rgb.jpg' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present

So the cropper examples are broken for anyone running the docs locally. The file exists at `apps/docs/public/assets/mittwald_logo_rgb.jpg`; the absolute URL is presumably deliberate so the shown code stays copy-pasteable, which is worth preserving — the `crossOrigin` attribute or a same-origin path for the cropper examples specifically would both work.

## Suggested split

1. Handle the load failure in `ImageCropper` (error state + a way for the app to observe it) — the actual bug.
2. Make the four cropper examples load locally.

Found while clearing the warnings in mittwald/flow#3019.

Issue #3017, opened 2026-08-31
react-live-ssr transpiles with the classic JSX runtime — 104 React warnings across the docs site


mfal:
Every docs page with a live code example logs:

> Your app (or one of its dependencies) is using an outdated JSX transform. Update to the modern JSX transform for faster performance.

104 of the 113 docs pages. It is the single loudest warning in the docs app.

## Cause

`@mfalkenberg/react-live-ssr@4.1.7` transpiles example code with sucrase and leaves `jsxRuntime` at its default. Sucrase's classic mode emits `React.createElement(type, { __self, __source, … })`, and React 19 warns exactly on a config that carries `__self` without `key` (`react.development.js`, `createElement`).

`dist/index.mjs`:

```js
return (code) => _transform(code, { transforms }).code;
```

## Verified fix

Adding `production: true` makes sucrase drop `__self`/`__source` and the warning disappears — confirmed by patching the installed copy and reloading the docs:

```js
return (code) => _transform(code, { transforms, production: true }).code;
```

`jsxRuntime: "automatic"` would also silence it but is riskier: react-live evaluates the transpiled code with only `React` in scope, so `_jsx` would need a runtime import the sandbox does not provide. `production: true` keeps the classic call shape.

## Getting it into the repo

The fix lives upstream in `@mfalkenberg/react-live-ssr`. A release cannot be consumed immediately — pnpm's `minimumReleaseAge` is one week and exempts only `@mittwald/*` — so a `patches/` entry is the bridge if we want the docs console clean before then.

Found while clearing the warnings in mittwald/flow#3019.

Issue #3015, opened 2026-08-31
Missing accessible names: DonutChart, List, ListItem and several stories warn in every dev run


mfal:
50 react-aria warnings across the docs site and Storybook, all variations of "this element has no accessible name". Each one is a real a11y gap, not console noise.

## `If you do not provide a visible label, you must specify an aria-label or aria-labelledby attribute for accessibility`

Stack traces put the warning inside react-aria's own components, reached from ours:

| Origin | Where it shows |
| --- | --- |
| `DonutChart` renders react-aria `ProgressBar` with no label | `data-visualisation-donutchart--default`, `--with-unit`, `--with-text-value` |
| `Slider` | `form-controls-slider--default`, `--with-format-options`, `04-components/form-controls/slider` |
| `Select` | `form-controls-select--with-country-options`, `--with-country-options-and-custom-sort` |
| `Modal` | `overlays-modal--long-content` |
| `ProgressBar` | `04-components/status/progress-bar` |

The docs examples for Slider and ProgressBar *do* pass a `<Label>`, so the warning comes from an inner react-aria element that never receives it — worth checking whether the label is forwarded, not just adding labels at the call site.

`DonutChart` has no label API at all today. Deciding what its `ProgressBar` should be called (or whether it should be `aria-hidden` with the value exposed elsewhere) is a UX question.

## `An aria-label or aria-labelledby prop is required for accessibility`

15 occurrences, all from react-aria's `GridListInner` — i.e. `List` rendering a `GridList` with no accessible name:

`structure-list-listitem--with-content`, `--with-action-group`, `--with-custom-tile-max-width`, `--with-checkbox`, `--with-column-layout`

## `A textValue prop is required for <GridListItem> elements with non-plain text children`

33 occurrences — the largest single group. Type-to-select does not work on these items.

`structure-list--with-summary`, `--infinite-scroll`, `--with-accordion`, `structure-list-edge-cases--custom-sorting`

## Scope note

Fixing this properly needs decisions the code cannot make: any label that ships inside a component becomes UI text and belongs in `locales/de-DE.locale.json` + `locales/en-US.locale.json`, which follows product voice. Story-level labels are free.

Found while clearing the warnings in mittwald/flow#3019.

mfal:
A second console sweep — all 443 Storybook stories, loaded **and** clicked through (up to 12 interactions each, overlays opened), two full passes with identical results — confirms this issue's findings and adds three origins it does not list yet.

[...]

Issue #3014, opened 2026-08-31
Two react-aria warnings point at real component bugs: uncontrolled→controlled and PressResponder


mfal:
A console sweep over the docs site, the remote-dom demo and every Storybook story surfaced two warnings that are not cosmetic — both indicate a component doing something react-aria considers wrong.

## `WARN: A component changed from uncontrolled to controlled.`

14 occurrences. React/react-aria fires this when a `selectedKey`-style prop arrives `undefined` on the first render and defined afterwards.

Affected:

- `structure-tabs--default`, `structure-tabs--collapsed`
- `04-components/structure/tabs`
- `04-components/form-controls/segmented-control`
- `03-patterns/01-patterns/detail-page`
- `02-foundations/03-content-guidelines/02-informationskonzept`

The `Tabs` / `SegmentedControl` selection state is the common factor. Whatever computes the initial selected key returns `undefined` for one render.

## `A PressResponder was rendered without a pressable child.`

8 occurrences. A `PressResponder` wraps something that does not call `usePress`, so the press handling is silently dropped — the wrapped element is not actually pressable.

Affected:

- `structure-section--multiple-sections`
- `03-patterns/01-patterns/detail-page` (6×)
- remote-dom demo `/host/context-menu`

`Section` and `ContextMenuTrigger` are the components in common.

## Why it matters

Both are dev-only messages, but each describes a real defect: the first means a controlled/uncontrolled state transition React explicitly warns about, the second means a press interaction that does not reach its target. Neither is a third-party quirk.

## Reproducing

Run the app and read the browser console, or sweep with Playwright:

```
pnpm nx dev components
```

then load `iframe.html?id=structure-tabs--default&viewMode=story` and `iframe.html?id=structure-section--multiple-sections&viewMode=story`.

Found while clearing the warnings in mittwald/flow#3019 — that PR fixed the build, test and dev-server warnings and left these two for their own investigation.

mfal:
Beide Hälften sind gemerged:

- #3025 — `PressResponder`: Overlay-Trigger-Buttons bleiben in ihrem Trigger
- #3048 — uncontrolled→controlled: `Tabs` hält die Auswahl ab dem ersten Render kontrolliert

#3048 hielt fest, dass keiner der beiden PRs das Issue allein schließt. Jetzt sind beide drin — geschlossen.

PR #3019, opened 2026-08-31
fix: unbreak the Storybook preview and clear the build, test and dev-server warnings

mfal:
## Storybook — superseded by #3010

While this was in review, #3010 landed the same one-line fix on `main`, so the merge kept the identical import and this PR now only adds the comment explaining it. The diagnosis for the record:

**Storybook's preview never rendered.** `.storybook/preview.tsx` imported `@teispace/next-themes` from the package root; that entry pulls in `next/navigation`, whose Next internals touch `process` at module scope. The preview is plain Vite with no Next runtime, so evaluating the dependency threw `ReferenceError: process is not defined` and the iframe stayed on the loading spinner — `pnpm nx dev components` was unusable. The `/client` entry has the same API for generic React.

## Warnings cleared

| Was | Now |
| --- | --- |
| ~60 lines of Vite `configLoader: 'native'` notices per test run | Every config import carries its extension (`./vite.config.ts`), points at a file instead of a directory (`../core/src/index.ts`), and declares the JSON import attribute. `allowImportingTsExtensions` added to the seven `tsconfig.vite.json` that lacked it |
| Rolldown `[PLUGIN_TIMINGS]` on the slower library builds | New shared `libraryBuildChecks` in `packages/core` turns the check off in all eight library builds. The hint only reports that unplugin-dts owns ~90% of plugin time — the declarations have to be generated, so there is no action behind it |
| `next dev` left a dirty tree on every start | `agentRules: false`. Next writes its managed AGENTS.md block unwrapped, prettier rewraps it at 80 columns, and the two overwrite each other forever — so every dev run dirtied the tree, and committing the block breaks `format:check` on push. The Next 16 pointer the block carried is now a hand-written line in both apps' AGENTS.md |

`pnpm test` and `pnpm build` are warning-free; `pnpm lint` has no new findings.

## Console findings

Swept every page with Playwright — 113 docs pages, 27 remote-dom-demo pages, 443 stories — and fixed what was mechanical:

- MarkdownEditor's `customTools` example rendered without an accessible name
- the typography page had `[Link]()`, an empty `href` React errors on
- the demo still called two deprecated APIs: `CartesianGrid` (→ `ChartGrid`) and `generatePasswordCreationFieldValidation` (→ `usePasswordCreationFieldValidation`)

## Follow-ups filed

The sweep found more than belongs in one PR:

- #3014 — two react-aria warnings that point at real component bugs: `uncontrolled → controlled` in `Tabs`/`SegmentedControl`, and `PressResponder` without a pressable child in `Section`/`ContextMenuTrigger`
- #3015 — 50 missing accessible names (`DonutChart`'s inner `ProgressBar`, `List`/`GridList`, `ListItem` `textValue`, plus Slider/Select/Modal stories). Needs UX input, since component-level labels become `locales/*` text
- #3017 — `@mfalkenberg/react-live-ssr` transpiles with sucrase's classic JSX runtime, producing 104 React warnings across the docs. Verified one-line fix (`production: true`)
- #3018 — `ImageCropper` turns a failed image load into an uncaught rejection, and its docs examples cannot load locally (CORS on the canvas read)

Left alone deliberately:

- recharts' `width(0) and height(0)` on the chart docs page. `.preview` in `LiveCodeEditor` is 48px wide at mount (its own padding — it is a centering column flex container), so everything inside measures 0 before layout settles. Transient and self-correcting, and not reachable through recharts props
- three `IconButton`/`ariaLabel` deprecations from `storybook-addon-rtl@3.0.2` — upstream
- the `SegmentedControl` deprecation warnings on its own docs page and stories, and `/releases` logging a GitHub 403 without a token — both by design

## Note on `pnpm install`

Nothing to fix there: a fresh install into an isolated copy of the manifests is completely warning-free on pnpm 11.19.0. The message that looks like an install warning is the `corepack pnpm` version mismatch on dependency-graph targets, which AGENTS.md already documents as a footgun — bare `pnpm nx …` avoids it.


ins0 (dismissed):

mfal:

mfal:

mfal:

ins0 (approved):

PR #3024, opened 2026-08-31
fix(ImageCropper): handle a failed image load instead of throwing

### LINKED ISSUE #3018: ImageCropper turns a failed image load into an uncaught rejection; its docs examples cannot load locally

mfal:
Fixes #3018.

## The defect

`getCroppedImageFile`'s `image.onerror` rejected and nobody caught it. The crop runs automatically right after the media loads, so **any** cropper whose image cannot be read into a canvas threw `Uncaught (in promise) Error: Failed to load image` on page load — with no error state and no way for the app to react. A cropper whose image never loaded at all had no signal either.

Two further paths left the promise **pending forever**: `toBlob` returning no blob, and a `SecurityError` from a tainted canvas. Both now reject.

## API — additive only

`ImageCropper` is `@flr-generate all`, so its props are a contract with extension developers. Nothing renamed or removed, so no `useWarnDeprecation` and no `MIGRATION.md` entry.

**`onError?: (error: ImageCropperError) => void`** — `{ reason: "load" | "crop", message: string }`, generated as a remote **event**. The payload is a plain object and deliberately **not** an `Error`: `FlowThreadSerialization` spreads unknown objects with `{...val}`, and `Error.message`/`stack` are non-enumerable — an `Error` would arrive in an extension as `{}`.

**`errorView?: ReactNode`** — generated as a **slot** (`remoteSlots: ["errorView"]`), the only correct shape for a prop carrying rendered output; a React element does not survive `postMessage` and would drop the whole mutation batch. Modelled on `CartesianChart`'s `emptyView`, but with a default: without `errorView` the component renders `IllustratedMessage color="danger"` + `IconDanger` + `Text`, following the error-handling pattern page.

On either failure the error view replaces the crop area inside the existing bordered box and the zoom slider is dropped — there is nothing to zoom. State resets when `image` changes, and a ref guarantees at most one `onError` per image. `ImageCropperError` is exported from the component barrel.

## New UI text — needs UX review

One string, `packages/components/src/components/ImageCropper/locales/`, key `error.text`:

- **de-DE:** `Das Bild konnte nicht geladen werden.`
- **en-US:** `The image could not be loaded.`

Shown for both `reason: "load"` and `reason: "crop"` — from the user's side the image is unusable either way.

## Docs examples

The four examples now use `/assets/mittwald_logo_rgb.jpg` instead of the cross-origin `https://flow.mittwald.de/assets/…`. Same file, same-origin, works locally and deployed; `basePath` is empty in both `deploy-main.yml` and `build-previews.yml`, so the bare path is safe.

Copy-pasteability is not the loss it would be elsewhere: for **this** component the cross-origin URL never worked for a consumer either — react-easy-crop reads the image into a canvas, so a pasted example displayed the image and then silently failed to produce a crop.

Two new examples demonstrate the error state and a custom `errorView`, plus a demo page in `apps/remote-dom-demo` for the remote path.

## Verified

- Playwright on `/04-components/upload/image-cropper`: 6 croppers, 4 images rendered, 2 error views, **0 page errors, 0 CORS / `ERR_FAILED`**. The remaining 404 is the deliberately broken image in the new error examples, and the remaining console error is that example's own `console.error` — i.e. the callback firing.
- `ImageCropper.browser.test.tsx`, 4 tests: load failure renders the error view, reports `reason: "load"` and drops the slider; `errorView` replaces the default; crop failure reports `reason: "crop"` **and asserts no `unhandledrejection`** — the exact reported defect; `getCroppedImageFile` rejects rather than hanging.
- `pnpm lint` 0 errors, `pnpm test` green, `git diff --exit-code` clean after `pnpm build`. Regenerated and committed: `RemoteImageCropperElement.ts`, `auto-generated/ImageCropper.ts`, `ImageCropper.module.d.scss.ts`; both props are in `doc-properties.json`.

## Visual baselines still needed

Two scenarios were added to `src/tests/visual/ImageCropper.browser.test.tsx` ("load error", "custom error view") and the visual suite was **not** run locally — three agents were building on this machine, and macOS writes only `-darwin` snapshots anyway. So the `visual` shards will fail on missing `-linux` baselines. I'll read the shard output first and only then apply `update-screenshots`, since that label runs `--update` over the whole suite and commits whatever renders at that moment.

**Version gate to check before merge:** the `errorView` scenario is gated `below: "1.0.16"` because `main` moved to the 1.0.15 bump mid-work. If this ships under a different version, that string needs updating.

## Open UX questions

- **No retry affordance.** The pattern page asks for a way out ("Erneut laden"), but a retry needs cache-busting and the consumer owns the image. `errorView` covers it for now.
- **No `Heading` in the default error view** — a 200–300px box is tight, so it is icon plus one line, unlike the pattern page's large-area example.
- **No live region** on the error state; it replaces content rather than announcing itself.
- **`crossOrigin` deliberately not set** on the displayed image. It would surface a CORS problem a beat earlier, but would turn today's "displays, cannot crop" into an immediate error for existing consumers — and since the crop fires right after load, the error appears within about a frame either way.


mfal:
Visual shards report exactly the expected shape: **only** the four new `ImageCropper` scenarios fail (two scenarios × Local/Remote), with 112 other tests green in that shard. They have no `-linux` baseline yet — nothing existing regressed.

Applying `update-screenshots` on that basis, since the suite is otherwise green. The label runs `--update` over the whole suite and `git add -A`, so I'll read the resulting commit's file list before this merges: anything beyond the four new `ImageCropper` PNGs would mean a flaky scenario got its current frame committed as truth, and would have to come back out.

packages/remote-react-components/src/tests/visual/__screenshots__/ImageCropper.browser.test.tsx/ImageCropper-custom-error-view-firefox-linux.png
ins0, line 1: @mfal this looks wrong from the non existing spacing - test issue?

mfal: Test issue — confirmed. The `IllustratedMessage` (icon + heading + text) does not fit the 200×200 box: the heading wrapped to two lines, the text to two more, and `.errorViewContainer` (`height: 100%` + `overflow: hidden`) clipped the last line against the border. Sized the scenario to 320×240 (1242de9af), where heading and text each fit one line; re-running `update-screenshots` for the Linux baselines. The clipping itself is correct for a fixed-size cropper box, so no component change.

ins0 (approved):

PR #3038, opened 2026-08-31
docs: an agent-driven upgrade prompt on the versioning page

mfal:
Two copy-to-clipboard prompts on the versioning page: one that drives a coding
agent through a Flow upgrade, and — separately, opt-in by copying it — one that
turns the run into feedback for us.

One file, `apps/docs/src/content/01-get-started/versioning/index.mdx`. No code, no
generator, no new component. Docs-only, so the merge publishes nothing.

Split out of a larger branch. Two companion PRs carry the code the second prompt
already produced findings about:

- mittwald/flow#3040 — `fix(codemods)`, to `main`
- mittwald/flow#3041 — `feat(codemods)`, to `next`

## Why the prompt points at a command, not at docs

Unlike the Next.js prompt this is modelled on, it names
`npx @mittwald/flow-codemods@latest list <revision> --json`. That already ships
the catalogue at the resolved version, carrying `apply`, `action` and `catchUp`
per entry — offline and machine-readable. Next.js needs "version-matched docs"
because their migration guide is a website; ours is in the package. Links in a
copied prompt would break outside the docs site, or pin to today's content rather
than the project's version.

## What the upgrade prompt encodes

**Which line to go to.** It runs `list patch|minor|major --json` first and shows
what actually resolved — target version, codemod/by-hand split — then asks. A
relative revision only resolves if such a release exists: from a prerelease
`patch` and `minor` exit non-zero, so an exact version or dist-tag is offered as
an equal option rather than a fallback.

**Every migration gets checked against the code, `action: "codemod"` included.**
Transforms are scoped narrowly on purpose and decline what they cannot decide
from the source, without reporting an error. The prompt points the agent at the
`upgrade` report's own admissions — changed no file, declined every file,
processed nothing — and adds that "no files were processed. Is the path right?"
is also what a crashed codemod looks like, so check stderr before believing the
path.

**Monorepos.** `resolveRange` reads exactly one manifest and walks no workspace,
so a single root run leaves every other package behind. The agent discovers where
Flow is declared first (via `git ls-files "*package.json"`, so `node_modules` and
build output like `.output` are excluded by construction), then runs once per
package with the resolved **exact** version — a relative revision resolves per
package and would spread them apart — committing between runs, because the
dirty-tree guard reads `git status` repo-wide.

**Install failures.** Beyond the publishing-window case, the prompt now covers
the general one: do not wait and retry, the manifest is already rewritten, read
the actual error and finish the install with the project's own package manager.

## The second prompt: feedback, opt-in

A separate block below the first. Copying it is the opt-in; not copying it is the
entire opt-out, so nobody answers a question about someone else's tooling in the
middle of their own upgrade. Pasted into the same session, the agent still has
the run in context.

It reports two levels. **The prompt itself** — a step in the wrong order, an
assumption the project does not meet, two instructions pulling apart, a situation
it says nothing about, an instruction that cost work and changed nothing. Nobody
sends this today because there is nowhere to send it. **The catalogue** — an
`apply` that was not executable without guessing, a codemod that declined what it
could have decided, a manual entry that is mechanically decidable, and the most
valuable class: a break the checks surfaced that no entry covers.

Three constraints: judge from what happened, not from re-reading the text (a
plausible critique nobody hit is worse than none, and this is the one channel
nothing can be checked against); the agent files nothing and says what the reader
can do with it; every code sample reduced to the smallest anonymised shape, since
the destination is a public tracker.

## It already worked

An agent ran the first prompt end to end on a Yarn 4 monorepo from
`0.2.0-alpha.584` to `1.1.0` and returned the second prompt's report. Six of its
findings were factual errors in the prompt — corrected here, each verified against
the code first. The largest: the prompt told the agent to read each entry's
`body`, and there is no `body` (`CatalogEntry` is `Omit<MigrationEntry, "body">`,
and the generated module is typed the same way), so `--json` never carries one.

## Verification

- `test:links` 30/30, `test:unit` 14/14, `prettier --check` clean
- Rendered locally: two separate copyable `CodeBlock`s in order, the `###`
  subsection at the same level as its neighbours, the tracker link resolves, no
  console or server errors
- Prompts checked at the source, not in the DOM — CodeMirror virtualises, so only
  ~43 lines are ever rendered. 122 and 57 lines, max width 80, no nested fence,
  no German text inside either. Prettier does not reflow fence content, so that
  wrapping is maintained by hand.

### DIFF
--- a/apps/docs/src/content/get-started/upgrades/index.mdx
+++ b/apps/docs/src/content/get-started/upgrades/index.mdx
@@ -12,9 +12,9 @@ description:
 Jede Änderung, die eine Anpassung in deinem Code erfordert, steht mit
 Vorher-Nachher-Beispiel im Migrationsleitfaden des betroffenen Packages:
 
-- [Migrationsleitfaden `@mittwald/flow-react-components`](https://github.com/mittwald/flow/blob/main/packages/components/MIGRATION.md)
+- [Migrationsleitfaden @mittwald/flow-react-components](https://github.com/mittwald/flow/blob/main/packages/components/MIGRATION.md)
   – deckt auch `@mittwald/flow-remote-react-components` ab
-- [Migrationsleitfaden `@mittwald/ext-bridge`](https://github.com/mittwald/flow/blob/main/packages/ext-bridge/MIGRATION.md)
+- [Migrationsleitfaden @mittwald/ext-bridge](https://github.com/mittwald/flow/blob/main/packages/ext-bridge/MIGRATION.md)
 
 Die Einträge sind nach Version absteigend sortiert und nennen jeweils die
 Version, ab der die Änderung greift. Suche die Version, von der du kommst, und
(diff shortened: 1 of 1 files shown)


apps/docs/src/content/get-started/upgrades/index.mdx
Lisa18289: das war zwar schon vor dem mr drin, aber ich find diese Links mit InlineCode schauen gar nicht schön aus, vlt lieber ohne InlineCode darstellen?

mfal: Guter Fang – und du hattest doppelt recht: das waren die einzigen zwei Links in den gesamten Docs mit InlineCode im Link-Text. Jeder andere Link labelt sich in Prosa. Ist in 7ea449e raus, die Package-Namen außerhalb der Links behalten ihr Code-Format.

Lisa18289 (approved):

Lisa18289 (approved):

PR #3042, opened 2026-08-31
fix(codemods): the SegmentedControl -> RadioGroup direction is a rename

mfal:
`segmented-control-deprecated`'s `apply` ends with **"this is a structural change,
not a rename"**. That is only true of one of the two directions, and the wrong
half is the one that sets the reader's expectation — it makes them price a
prop-compatible swap as a rewrite, on an entry that is `action: "manual"`.

Reported from a real unattended upgrade run (the feedback prompt in
mittwald/flow#3038). Verified against the types before changing anything:

```ts
SegmentedControlProps extends Omit<Aria.RadioGroupProps, "children" | "ref">
RadioGroupProps       extends Omit<Aria.RadioGroupProps, "children">
SegmentProps          extends Omit<Aria.RadioProps, "children">
RadioProps            extends Omit<Aria.RadioProps, "children">
RadioButtonProps    = RadioProps
```

`SegmentProps` and `RadioProps` are the same type. So towards `RadioGroup` the
migration **is** a rename: `SegmentedControl` → `RadioGroup`, `Segment` →
`RadioButton`, with `value` / `defaultValue` / `onChange` and a `Label` child
carrying over untouched. Only `containerBreakpointSize` has no counterpart, which
the entry already covered. Towards `Tabs` it is genuinely structural.

## Four further gaps, each checked in the code

- **`RadioButton` was never mentioned.** It takes exactly `Segment`'s props, so it
  is as much a drop-in as `Radio`. Which of the two to use is a design call, and
  the review made it: **always `RadioButton`, never the plain `Radio`.** The entry
  names only `RadioButton` — in `apply`, in the prose, in the diff example, and in
  `Segment`'s `@deprecated`. `Radio` survives in exactly one sentence: the one
  saying not to use it.

  What stays is the warning that `RadioButton` does *not* reproduce the joined
  row — that came from `Segment`'s own `flex: 1` and collapsed borders. The
  appearance changes, and that should not be a surprise after the rename.

- **The Tabs direction renames the state props too**: `selectedKey` /
  `defaultSelectedKey` / `onSelectionChange` rather than `value` / `defaultValue` /
  `onChange`. Nothing carries over by name there.

- **`Tabs` has no `Label` slot**, so the group label needs somewhere else to go.
  Three places, since #3045: the surrounding `Heading` when the group wants a
  visible title, `aria-label` on `Tabs` when it does not, or nowhere when the
  heading already names the group. A `Label`'s presence is *not* a signal the
  usage was a `RadioGroup` case, since `SegmentedControl` required a field label
  even for content switchers.

- **Inactive tabs stay mounted.** `Tab` passes `shouldForceMount` to
  `Aria.TabPanel` and wraps `Content`/`Section` in `<Activity isActive={…}>`, so
  moving panels that register form fields into tabs does not change when those
  fields mount. That was the reader's main reason to hesitate, and the entry now
  answers it.

## The "no codemod" reason was wrong too

It read "both replacements change the structure, and picking the right one is a
decision per usage, not a rename". The *edit* is mechanical in the `RadioGroup`
direction; what is undecidable is the **choice**, because a value-setting usage
and a content switcher look alike at the call site. The entry says that, and adds
that a find-and-replace covers most of it once the decision is made.

## Also

`Segment`'s own `@deprecated` named only `Radio`; it now names both. Confirmed
that changes no generated artefact — `build:status-registry` stores the lifecycle
status, not the deprecation text, and both it and `build:remote-components`
produced no diff.

## Verification

- `test:unit` codemods 252/252, `test:compile` for **codemods and components**
  clean, `pnpm lint` exit 0, `prettier --check` clean repo-wide
- `MIGRATION.md` and `migrations.generated.ts` regenerated and committed
- The RadioGroup docs link the entry adds resolves to an existing content
  directory

## Merge after #3045

The `aria-label` option above only works once #3045 forwards the prop from `Tabs`
to the tab list. Before that, `Tabs` accepts `aria-label` and drops it on a
role-less wrapper `div`, leaving `role="tablist"` unnamed — so merging this first
would ship a `MIGRATION.md` describing something that does not work yet.

Split out of the `feat(codemods)` PR to `next` on request, so a correction to an
entry that currently states something false ships on the next patch rather than
waiting for a minor. Verified that this branch merges into that one with no
conflict.


packages/codemods/src/migrations/segmented-control-deprecated/entry.md
```diff
@@ -4,10 +4,20 @@ title: SegmentedControl deprecated
...
 action: manual
 remotePackage: true
-apply:
+apply: >-
   Replace `SegmentedControl` with `Tabs` when the selection switches displayed
-  content, or with `RadioGroup` when it sets a value. Pick per usage — this is a
-  structural change, not a rename.
+  content, or with `RadioGroup` when it sets a value. Pick per usage. The two
+  directions cost very different amounts of work. Towards `RadioGroup` it is a
+  prop-compatible rename: `SegmentedControl` → `RadioGroup`, `Segment` → `Radio`
```
Lisa18289: wir wollen beim Umbau von SegmentedControl zu RadioGroup immer den RadioButton verwenden, nicht das einfache Radio

packages/codemods/src/migrations/segmented-control-deprecated/entry.md
```diff
@@ -39,8 +49,15 @@ tracked the selection both go away.
 ```
 
 **Setting a value.** The selection feeds a form or a setting. Use a
-`RadioGroup`: each `Segment` becomes a `Radio`, everything else — `Label`,
-`FieldDescription`, `FieldError`, the React Hook Form binding — stays as it is.
+`RadioGroup`: each `Segment` becomes a `Radio` or a `RadioButton`, everything
```
Lisa18289: hier auch, nur "RadioButton", und auch an allen weiteren Stellen

packages/codemods/src/migrations/segmented-control-deprecated/entry.md
```diff
@@ -4,10 +4,20 @@ title: SegmentedControl deprecated
...
-  structural change, not a rename.
+  content, or with `RadioGroup` when it sets a value. Pick per usage. The two
+  directions cost very different amounts of work. Towards `RadioGroup` it is a
+  prop-compatible rename: `SegmentedControl` → `RadioGroup`, `Segment` → `Radio`
+  or `RadioButton` (identical props to `Segment`; `RadioButton` is the boxed
+  one, though neither reproduces the joined row), and
+  `value`/`defaultValue`/`onChange` and a `Label` child all carry over. Only
+  `containerBreakpointSize` has no counterpart. Towards `Tabs` it is structural:
+  the state props are `selectedKey`/`defaultSelectedKey` rather than
+  `value`/`defaultValue`, there is no `Label` slot (the group label moves to the
```
Lisa18289: hier sollte dann aria-label erwähnt werden sobald die Tabs das unterstützen

packages/codemods/src/migrations/segmented-control-deprecated/entry.md
mfal: Umgesetzt in 94ddfea. Die Entscheidung steht einem Migrationseintrag nicht zu — ihr habt sie getroffen, jetzt steht sie drin.

`apply` nennt nur noch `SegmentedControl` → `RadioGroup` und `Segment` → `RadioButton`, mit dem expliziten „nicht `Radio`“. Der Absatz, der die Wahl zwischen beiden eröffnet hat, ist weg.

Stehen geblieben ist nur der Hinweis, dass `RadioButton` die zusammenhängende Zeile des SegmentedControl nicht nachbildet — die kam aus `Segment`s eigenem `flex: 1` und den zusammengezogenen Rahmen. Die Optik ändert sich also, und das sollte niemand erst nach dem Umbau merken.

packages/codemods/src/migrations/segmented-control-deprecated/entry.md
mfal: Ja, überall umgestellt:

- Fließtext „jedes `Segment` wird ein `RadioButton`“
- Diff-Beispiel: `<RadioButton value="debit">Lastschrift</RadioButton>`
- Typ-Absatz: `SegmentProps` und `RadioButtonProps` sind dasselbe (`RadioButtonProps = RadioProps`, und `RadioProps` ist identisch zu `SegmentProps`)
- `Segment`s `@deprecated`-JSDoc in `Segment.tsx`

`Radio` kommt jetzt genau einmal vor: in dem Satz, der sagt, dass man es hier nicht nimmt. `MIGRATION.md` und `migrations.generated.ts` sind neu generiert.

packages/codemods/src/migrations/segmented-control-deprecated/entry.md
mfal: Das geht heute noch nicht — `Tabs` kann gar nicht beschriftet werden.

`TabsProps` reicht zwar `Aria.TabsProps` durch, aber in react-aria gehört `aria-label` an die `TabList`. Flows `TabList` ist eine interne Komponente ohne Props-Durchreichung: sie rendert `<Aria.TabList className ref>` und sonst nichts. Es führt also kein Weg von außen an das Element, das das Label tragen müsste. Ein Issue dafür gibt es auch noch nicht.

Deshalb lasse ich diesen Thread offen und den Absatz vorerst so, wie er ist (Label wandert in die umgebende `Heading` oder entfällt). `MIGRATION.md` beschreibt eine veröffentlichte Version — ein `aria-label`, das es nicht gibt, wäre dort genau der Fehler, den dieser PR gerade behebt.

Soll ich ein Follow-up-Issue für das Label an `Tabs` anlegen? Dann ergänze ich den Absatz, sobald es steht.

mfal: Nachtrag: erledigt in 537c96c — #3045 schließt genau diese Lücke und ist gleich drin.

Damit hat der `Label`-Text drei Ziele statt zwei, und der Absatz listet sie jetzt auch als solche:

- die umgebende `Heading`, wenn die Gruppe einen sichtbaren Titel haben soll
- `aria-label` an `Tabs`, wenn nicht — benennt die Tab-Liste für Screenreader, ohne den Namen zu zeigen
- gar nichts, wenn die umgebende Überschrift die Gruppe schon benennt

Auch im `apply`-Feld, denn für einen `action: "manual"`-Eintrag ist das die eigentliche Anweisung.

Eine Reihenfolge-Abhängigkeit bleibt: #3045 muss **vor** diesem PR gemergt werden. Bis dahin nimmt `Tabs` `aria-label` zwar entgegen, setzt es aber auf ein `div` ohne Rolle — die `role="tablist"` bliebe unbenannt, und `MIGRATION.md` würde etwas beschreiben, das noch nicht funktioniert.

Lisa18289 (approved):

PR #3045, opened 2026-09-01
fix(Tabs): forward aria-label to the tab list

mfal:
`Tabs` accepted `aria-label` and dropped it on an element that has no role. The
tab list — the thing that needs a name — never saw it.

`TabsProps extends Omit<Aria.TabsProps, "children">`, and `AriaTabListProps`
carries `AriaLabelingProps`, so the prop type-checked and even showed up in the
generated properties table. `Aria.Tabs` runs `filterDOMProps(props, {global:
true})` and puts the result on its wrapper `<div>`; `useTabList` reads labelling
only from `Aria.TabList`'s own props. Result: `<Tabs aria-label="…">` rendered
the attribute on a role-less div and left `role="tablist"` unnamed.

Both `aria-label` and `aria-labelledby` now travel to `TabList` and onto
`Aria.TabList`. `aria-describedby` and `aria-details` stay on the wrapper, where
react-aria puts them — they describe, they don't name.

## Why it matters here

`Tabs` has no `Label` slot; the tab list names itself through its tab titles. So
when the surrounding heading doesn't already name the group, `aria-label` is the
only way — and it was the one that didn't work. This is the gap a
`SegmentedControl` → `Tabs` migration walks into: the `<Label>` of a
`SegmentedControl` has nowhere to go (see mittwald/flow#3042).

## What's in the change

- `Tabs.tsx` / `TabList.tsx` — forward both labelling props to `Aria.TabList`
- `TabsProps` — both declared explicitly with JSDoc. They were already listed in
  the properties table with react-aria's generic wording; now the table says
  what they name and that there is no `Label` slot
- `Tabs.browser.test.tsx` — new. `aria-label` and `aria-labelledby` each name
  the `tablist`. Verified red before the fix (4/4 failing across webkit and
  firefox), green after
- Docs: a **Beschriftung** section on the Tabs page, and `aria-label` in the
  Storybook meta args so every story ships a named tab list

## Verification

- `test:browser` components (webkit) — Tabs 4/4 pass; the same two tests fail
  4/4 with the forwarding removed
- `test:compile` components clean, `test:unit` components 253/253, `pnpm lint`
  exit 0
- `build:remote-components` produced no diff — `aria-label` and
  `aria-labelledby` are already remote properties of `flr-tabs`, so the remote
  path works without a protocol change
- No usage in this repo passed `aria-label` to `Tabs`, so nothing loses the
  attribute on the wrapper. No visual output changes; no visual scenario passes
  the prop, so no cross-version divergence

## Follow-up

Once this and mittwald/flow#3042 are both in, the `segmented-control-deprecated`
entry can name a third option for the group `Label` — today it says the label
moves to the surrounding `Heading` or goes away. Left out here to keep the two
PRs conflict-free.

### DIFF
--- a/apps/docs/src/content/components/structure/tabs/index.mdx
+++ b/apps/docs/src/content/components/structure/tabs/index.mdx
@@ -62,6 +62,29 @@ aktiv sein, legst du das über `defaultSelectedKey` fest.
 
 ---
 
+# Beschriftung
+
+Tabs haben keinen `Label`-Slot – die Tab-Leiste beschriftet sich über ihre
+Tab-Titel. Benennt die Überschrift der umgebenden `Section` die Gruppe bereits,
+reicht das. Wenn nicht, gib der Tab-Leiste über `aria-label` einen eigenen Namen
+– oder verweise mit `aria-labelledby` auf eine vorhandene Überschrift, damit der
+Name nur an einer Stelle gepflegt wird.
+
+```tsx
+<Tabs aria-label="Verbindungsart">
+  <Tab>
+    <TabTitle>Mit App verbinden</TabTitle>
+    App Content
+  </Tab>
+  <Tab>
+    <TabTitle>Mit Container verbinden</TabTitle>
+    Container Content
+  </Tab>
+</Tabs>
+```
+
+---
+
 # Reihenfolge der Tabs
 
 Werden innerhalb eines Tabs Inhalte nachgeladen, kann sich durch das Rerendering
--- a/apps/docs/src/content/components/structure/tabs/index.mdx
+++ b/apps/docs/src/content/components/structure/tabs/index.mdx
@@ -70,18 +70,7 @@ reicht das. Wenn nicht, gib der Tab-Leiste über `aria-label` einen eigenen Name
 – oder verweise mit `aria-labelledby` auf eine vorhandene Überschrift, damit der
 Name nur an einer Stelle gepflegt wird.
 
-```tsx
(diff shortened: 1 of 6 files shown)


apps/docs/src/content/components/structure/tabs/index.mdx
```diff
@@ -62,6 +62,29 @@ aktiv sein, legst du das über `defaultSelectedKey` fest.
...
 
+# Beschriftung
+
+Tabs haben keinen `Label`-Slot – die Tab-Leiste beschriftet sich über ihre
+Tab-Titel. Benennt die Überschrift der umgebenden `Section` die Gruppe bereits,
+reicht das. Wenn nicht, gib der Tab-Leiste über `aria-label` einen eigenen Namen
+– oder verweise mit `aria-labelledby` auf eine vorhandene Überschrift, damit der
+Name nur an einer Stelle gepflegt wird.
+
+```tsx
```
Lisa18289: vlt lieber ein LiveCodeEditor Beispiel? Haben wir ja sonst auch überall so

mfal: Umgestellt: das Snippet liegt jetzt als `examples/aria-label.tsx` und die Seite bindet es mit `<LiveCodeEditor example="aria-label" editorCollapsed />` ein — wie die übrigen Abschnitte. Lokal geprüft: das Beispiel rendert, und die Tab-Leiste trägt darin den Namen "Verbindungsart".

Lisa18289 (approved):

PR #3051, opened 2026-09-01
fix(Link): prevent text selection on icon element

TimAufdemKampe:


Lisa18289:
I can't reproduce the effect — as far as I can measure, the rule does nothing.

Dragging a selection over the link behaves identically with and without it in all three engines: same selected text, pixel-identical highlight. Which makes sense — the icon is an inline SVG with no text in it, so there's nothing to select there.

What does get selected is the zero-width joiner that LinkIcon renders next to the icon: copying a link with target="_blank" gives you mittwald.de plus an invisible U+200D. If that's what bothered you, the rule would have to go on .linkIcon (the wrapping <span>) instead of .icon (the svg).

Whichever way it goes, we'd like a test for it. A visual test won't help — selection isn't captured in a screenshot — so a *.browser.test.tsx asserting the computed style, or the selected text after a drag, is the way to cover this.

What did you actually observe? No issue is linked, so I might be testing the wrong thing.

Lisa18289 (dismissed):

Lisa18289 (approved):

PR #3053, opened 2026-09-01
fix(Button): treat a string label like a Text child

Lisa18289:
A `Button` with an icon and a label lost the spacing between them inside a `Link`. Digging into that surfaced the same root cause one level down, so this fixes both.

**`Link`** wrapped all `Button` children in a single `Text`, collapsing icon and label into one flex item — the button's `column-gap` had nothing to apply to.

**`Button`** only wrapped children in `Text` when they were a lone string, and `:has(.text)` was the only signal that a button contains text. So a raw string next to an icon matched nothing: `<Button><Icon />Label</Button>` got the icon-only padding (8px instead of 16px), and a plain pending button hid its label. Which branch a button took depended on how the label happened to be written — `Label` and `{x} Label` behaved differently.

The button now carries the marker itself, so a string is detected wherever it sits, including through the fragment `Link` passes down. That drops the `Text` wrapper from both components; `.content` keeps the two properties it contributed (`text-align`, `text-wrap-style`).

Also removes a dead selector: `&:where(:has(.text) .icon)` compiles to `.flow--button:where(:has(.flow--button--text) .flow--button--icon)` — a compound selector that can never match.

Baselines: three scenarios change on purpose — the icon+string padding, the plain pending label, and sub-pixel text rendering where a label is no longer inside a span. The darwin baselines are updated; the linux ones come from the `update-screenshots` label.


packages/components/src/lib/react/remote.ts
```diff
@@ -16,6 +22,24 @@ export function isRemoteComponentRendererProps(
   );
 }
 
+/** Whether the node renders text — a string, or its remote counterpart. */
+const isTextNode = (child: ReactNode): boolean =>
+  typeof child === "string" ||
+  (isValidElement(child) && isRemoteTextRenderProps(child.props));
```
mfal: Let `isTextNode` count numbers too.

This PR is the first to put the "is this child text" predicate into `remote.ts`, and two open PRs are about to consume it:

- #3070 carries a local `isText` in `Navigation.tsx` — same three branches, **including `number`**. Once it drops that helper for `containsTextChild`, a numeric label (`<Link>{42}</Link>`) would stop being wrapped and stop truncating. That is a regression #3070 does not have today.
- #3076 adds `extractTextFromChildren` here and needs the numeric branch as well (it imports `isNumber` from remeda for it).

So the primitive should land as the superset right away — otherwise the PR that consolidates the three has to widen it after the fact, and #3070 silently loses a case in between.

`typeof` over remeda's `isNumber` here keeps the change to these four lines; the imports stay as they are.

```suggestion
/** Whether the node renders text — a string, number, or remote counterpart. */
const isTextNode = (child: ReactNode): boolean =>
  typeof child === "string" ||
  typeof child === "number" ||
  (isValidElement(child) && isRemoteTextRenderProps(child.props));
```

No behavior change for this PR's own callers: `Button` and `Link` treat a numeric label as a label, which is what the CSS marker is for.

Lisa18289: Applied in f6eab9b27, with the `typeof` form as suggested — imports unchanged.

I checked both referenced PRs and the claims hold: #3070 does carry a local `isText` in `Navigation.tsx` with the `number` branch, and #3076 does add `extractTextFromChildren` to this file.

Worth noting it is not only a forward-compatibility fix — this PR needed it on its own. `<Button><Icon />{42}</Button>` matched no text marker and got the icon-only padding (8px instead of 16px), which is exactly the bug this PR fixes for strings. Verified in Storybook: `{42}` and `42` now both carry the marker and render at 16px, while the icon-only button stays at 8px.

Unit 253/253, browser 524/524, and the `Button` / `Link` visual scenarios still match their baselines.

mfal (approved):

PR #3068, opened 2026-09-02
chore(core): stop writing failure screenshots next to tests

Lisa18289:
Vitest's `screenshotFailures` defaults to on, and no `__screenshots__` directory is gitignored — they hold the committed visual baselines — so a red browser test drops a PNG that `git add -A` sweeps into the commit. Two such strays are on main already, in `List` and `MarkdownEditor`, referenced by no test.

`browser.screenshotDirectory` in the shared browser config moves the capture to a gitignored `.vitest-screenshots` instead of switching it off, so the evidence survives: the runner prints the path on failure. That covers every package that inherits the config, including future browser projects, so the local `screenshotFailures: false` overrides in `remote-react-components` become redundant.

For an ordinary browser test — click, assertion, timeout — that screenshot is the only visual evidence there is, and a firefox/webkit failure that only reproduces in CI cannot be recovered with a local `--browser.screenshotFailures`. `toMatchScreenshot` failures are unaffected either way: the runner already skips the failure screenshot for them, and reference, actual and diff go to the gitignored `.vitest-attachments`.

### The catch this pins

`screenshotDirectory` also feeds the default path of the `toMatchScreenshot` **baselines**, but with different semantics — a root of its own for `page.screenshot()`, a segment *inside* the test directory for `toMatchScreenshot`. Unpinned it would move the 954 committed baselines under `packages/remote-react-components/src/tests/visual/__screenshots__` into the gitignored directory, and the visual suite would re-baseline itself on the next run and stop catching anything — a worse failure mode than a few stray PNGs.

So `resolveScreenshotPath` spells out vitest's documented default with `__screenshots__` fixed. The `.gitignore` entry is bare, matching at any depth, because the directory resolves against the process cwd rather than the project root.

`PATTERNS.md` listed screenshot baselines as a pattern of the components package — the two strays were its only evidence. It points at the visual suite now.

### Verified

On vitest 4.1.11, both directions:

- a deliberately failing browser test in `components` wrote `packages/components/.vitest-screenshots/src/components/Button/…png`, git ignores it via `.gitignore:18`, and no `__screenshots__` directory appeared anywhere under `src`
- `Alert.browser.test.tsx --project=visual --browser.name=webkit` passed 6/6 against the existing baselines, which kept their checkout mtime — the pinned path resolves exactly where they already are

Test config, docs and deleted artifacts only — nothing to release.


mfal:
Not a blocker, just an alternative I'd like to put up for discussion: my worry with `screenshotFailures: false` is that we don't just stop the strays, we also stop seeing what went wrong. Vitest has an option that gets the first without paying the second.

### `browser.screenshotDirectory`

Verified against the installed vitest 4.1.11, `@vitest/browser/dist/index.js:645`:

```js
if (config.browser.screenshotDirectory) {
  return resolve(config.browser.screenshotDirectory, relative(config.root, dir), base, name);
}
return resolve(dir, "__screenshots__", base, name);
```

The automatic failure screenshot goes through exactly this path (`onTaskFinished` calls `page.screenshot()` without `path`), so pointing it at a gitignored root-level directory keeps the evidence and still gets nothing next to the tests.

### The catch, if we go this way

The same field also feeds the default path of the `toMatchScreenshot` **baselines** — but with different semantics: there it is a segment *inside* the test directory (`${root}/${testFileDirectory}/${screenshotDirectory}/${testFileName}/…`), whereas for `page.screenshot()` it is a root of its own. Set globally, it would move the 952 committed baselines under `packages/remote-react-components/src/tests/visual/__screenshots__` into a gitignored directory — the visual suite would silently re-baseline on the next run and stop catching anything. That's a worse failure mode than a few stray PNGs, so it needs pinning:

```ts
browser: {
  screenshotDirectory: ".vitest-screenshots",
  expect: {
    toMatchScreenshot: {
      resolveScreenshotPath: ({ root, testFileDirectory, testFileName, arg, browserName, platform, ext }) =>
        `${root}/${testFileDirectory}/__screenshots__/${testFileName}/${arg}-${browserName}-${platform}${ext}`,
    },
  },
},
```

Plus one `.vitest-screenshots` line in `.gitignore` — it matches at any depth, same as `.vitest-attachments`, which matters because `screenshotDirectory` resolves per project root.

### How much we actually lose without it

Less than the description suggests, but precisely where it hurts: on a `toMatchScreenshot` failure the runner skips the failure screenshot anyway (it checks `assertionName === "toMatchScreenshot"`), and reference/actual/diff already land in `.vitest-attachments`. What goes away is the only visual evidence for **ordinary** browser tests — click, assertion, timeout. That's the majority in `components`, and `--browser.screenshotFailures` only helps when you can reproduce locally; a flaky firefox/webkit failure in CI is gone.

### Side note

The problem keeps reproducing while this PR is open: `573f50928` on `fix/section-container-type-2655` picked up three more strays in `List/__screenshots__` (`does-not-blow-out-an-equal-width-grid-track--{list,table,tiles}-view--1.png`). They come in with that branch, so there is nothing for this PR to do about them — but they are a live argument for fixing the mechanism rather than only the two files.

_Edit: this section originally claimed those three were already on `main` and that this branch needed a rebase to catch them. Both wrong — I had mistaken a local feature-branch checkout for `main`. `main` has exactly the two strays this PR deletes, so the cleanup here is complete. Sorry for the noise._

mfal (dismissed):
See my latest comment

mfal:
Pushed the alternative from my comment above as 5731fb9, and updated the PR description to match — the body described `screenshotFailures: false`, which the branch no longer does.

What changed against the previous state of the branch:

- `screenshotFailures: false` → `browser.screenshotDirectory: ".vitest-screenshots"`, so a failing browser test still captures the page, just into a gitignored directory whose path the runner prints
- `browser.expect.toMatchScreenshot.resolveScreenshotPath` pins the baseline path to `__screenshots__`. This is the load-bearing part: `screenshotDirectory` feeds both paths with different semantics, and unpinned it would relocate the 954 committed baselines into the gitignored directory and let the visual suite re-baseline itself
- one `.vitest-screenshots` line in `.gitignore`, bare so it matches at any depth — the directory resolves against the process cwd, not the project root
- `PATTERNS.md` now points at `screenshotDirectory` instead of `screenshotFailures`

Everything else on the branch is unchanged: the two stray PNGs stay deleted, and the local `screenshotFailures: false` overrides in `remote-react-components` stay removed — redundant either way, and here they would also throw away the evidence.

Verified on vitest 4.1.11 in both directions: a deliberately failing browser test in `components` wrote `packages/components/.vitest-screenshots/src/components/Button/…png` with git ignoring it and no `__screenshots__` directory appearing under `src`; `Alert.browser.test.tsx --project=visual --browser.name=webkit` passed 6/6 against the existing baselines, which kept their checkout mtime, so the pinned path resolves exactly where they already sit.

@Lisa18289 — this is your PR, so overrule me freely if you prefer the simpler switch-off. @ins0 added as reviewer.

Note my earlier `CHANGES_REQUESTED` is still pending and blocks auto-merge (which is enabled on this PR); I left it for a human to resolve rather than approving my own commit.

ins0 (approved):

PR #3076, opened 2026-09-02
fix(Option): infer textValue from mixed children and keep value stable

### LINKED ISSUE #3028: Option with mixed children loses both textValue and value, falling back to an unstable generated key
An `<Option>` whose children are not a single text node loses both its `textValue` and its `value`. The a11y warning is the visible symptom; the unstable key is the real damage.

## Root cause

[`Option.tsx:24-25`](https://github.com/mittwald/flow/blob/main/packages/components/src/components/Option/Option.tsx#L24):

```ts
textValue = extractTextFromFirstChild(children),
value = textValue,
```

`extractTextFromFirstChild` ([`lib/react/remote.ts`](https://github.com/mittwald/flow/blob/main/packages/components/src/lib/react/remote.ts)) bails out immediately:

```ts
if (Children.count(children) !== 1) {
  return undefined;
}
```

So any `Option` with more than one child — text plus a `Badge`, an icon, anything — gets `textValue === undefined` **and** `value === undefined`.

## What that costs

react-aria then falls back to a generated key. From the `Select` Default story, which renders `<Option>Millennium Falcon <Badge>Latest</Badge></Option>` next to plain-text options:

| Option | key |
| --- | --- |
| `Millennium Falcon` + `<Badge>` | `react-aria-1` |
| `X-Wing` | `X-Wing` |
| `TIE Fighter` | `TIE Fighter` |

Three consequences:

1. The selected value for that option is `"react-aria-1"`, not `"Millennium Falcon"`. A form submits a meaningless key.
2. That key comes from a render-order-dependent counter, so `defaultSelectedKey` / controlled `selectedKey` cannot reliably target the option, and the key can shift when unrelated markup around it changes.
3. Type-to-select skips the option, which is what react-aria's warning is about: `A textValue prop is required for <ListBoxItem> elements with non-plain text children`.

Points 1 and 2 are silent — nothing in the console mentions them.

## Affected stories

`form-controls-select--default`, `form-controls-select-edge-cases--long-texts`, `form-controls-select-edge-cases--many-options`

The pattern is in the Default story itself, so it is the documented usage rather than an edge case.

## Not a duplicate

#3015 covers the same class of warning for `GridListItem` in `List`. This is `ListBoxItem` in `Option`/`Select` — a different component, a different inference helper, and it additionally corrupts `value`, which #3015 does not touch.

## Direction

Two separable decisions:

- **Inference.** Walk the children for text instead of requiring exactly one node, so the common "text + Badge" shape keeps a sensible `textValue`.
- **`value` defaulting to `textValue`.** Even with better inference this couples the form value to display text. Letting `value` silently become `undefined` is the part that should not survive — failing loudly, or decoupling the two, is worth deciding explicitly.

`extractTextFromFirstChild` also handles the remote-text case, so any change has to keep remote `Option`s working.

## Reproducing

```bash
pnpm nx dev components
```

Open `iframe.html?id=form-controls-select--default&viewMode=story`, open the select, and inspect the first option's `data-key` — it is `react-aria-1` while its siblings carry their text.

Found in a Playwright console sweep over all 443 stories (two full passes, identical results).

mfal:
## Root cause

`Option.tsx` defaulted `textValue = extractTextFromFirstChild(children)` and
`value = textValue`. `extractTextFromFirstChild` bails out on
`Children.count(children) !== 1`, so text next to any element left **both**
undefined and react-aria assigned the item a key off a render-order counter.

That key is the option's form value and what `defaultValue` / `selectedKey` has
to match. So `<Option>Millennium Falcon <Badge>Latest</Badge></Option>` — the
Select Default story — submitted `"react-aria-1"`, could not be preselected, and
its key shifted when unrelated markup around it changed. Only the missing
`textValue` warned; the key damage was silent.

## Decision 1 — inference walks the children

New helper `extractTextFromChildren` (`src/lib/react/remote.ts`) joins the text
among the children instead of demanding exactly one node.

**Only text that is a child itself counts.** Text inside an element child
belongs to that element, so the option above is `"Millennium Falcon"`, not
`"Millennium Falcon Latest"` — the Badge is decoration, not identity. Numbers
count, fragments are transparent, and `RemoteTextRenderer` elements count the
same as strings, which is what keeps remote Options working.

**`extractTextFromFirstChild` is untouched.** Its four other callers all depend
on the strict "exactly one text node" reading, and would break on a walk:
`Button` decides its icon-only layout by it, `Markdown` takes it as the markdown
source, `Initials` and `Truncate` operate on the whole content. Hence a new
helper rather than a changed one; both now carry JSDoc saying which question
they answer.

## Decision 2 — `value` keeps its default, but fails loudly

`value` still defaults to `textValue`. Decoupling them is the bigger breaking
change: every app that relies on text-as-value would lose its keys at once, and
that default is a contract with extension developers.

What does not survive is `value` silently becoming `undefined`. An `Option` with
no inferable text now logs

> An &lt;Option&gt; has no 'value' and none could be inferred from its children,
> so it falls back to a generated, render-order-dependent key. Pass a 'value' to
> give the option a stable key.

once per option, from an effect (so StrictMode does not double it). react-aria
already warns about the missing `textValue` in that situation but says nothing
about the key, which is the expensive half.

## Migration entry — yes, needed

`packages/codemods/src/migrations/option-value-inferred-from-mixed-children/`
(`action: none`, `remotePackage: true`). No prop changed shape, so this is a
changed *default*, not a breaking prop change — but the **selected key of an
existing app can change**, from `react-aria-N` to the option's text. Anywhere
that key was read back needs checking: a stored or server-side selection, a
`defaultValue` matched against it, a test asserting on it. Not mechanically
decidable from source, so no codemod.

`MIGRATION.md` and `migrations.generated.ts` regenerated and committed;
`guide.test.ts`'s pinned newest-entry id updated.

## Verification

Test-first. Reverting `Option.tsx` to the old inference fails 4 of the 6 new
browser tests, with `expected [ 'react-aria-1', 'X-Wing', … ]` — the issue's
exact symptom.

- `src/lib/react/remote.test.tsx` — 20 unit tests over both helpers, including
  the remote-text and fragment paths
- `Option/Option.browser.test.tsx` — pins the `data-key` list of the Select
  Default story shape, the emitted `onChange` value, `defaultValue` targeting,
  and the new console error
- `remote-react-components/src/tests/OptionKeys.browser.test.tsx` — the same
  keys in **Local and Remote**, the only guard for the remote-text path

Passed: `pnpm nx test:unit components` (35 files / 285),
`pnpm nx test:compile components`,
`pnpm nx test:browser components --browser.name=webkit` (39 / 268),
`pnpm nx test:browser remote-react-components --browser.name=webkit` (5 / 12),
`pnpm nx test:unit codemods` (30 / 309), `pnpm nx test:compile codemods`,
`pnpm lint` (0 errors), `pnpm build` (no further generated diff — props kept
their shape, so `view.ts` and the `auto-generated` dirs are unchanged).

No rendered output changes, so no visual test was touched; `run-visual-tests` is
set to confirm that.

## Decision 3 — `containsTextChild` folds into this walk

#3053 merged meanwhile and added `containsTextChild` to the same file, built on
its own `isTextNode` predicate and its own fragment recursion — the same
classification `joinTextChildren` does here, traversed twice.

It is now one line over the walk:

```ts
export const containsTextChild = (children: ReactNode): boolean =>
  extractTextFromChildren(children) !== undefined;
```

`isTextNode` is gone. One behavior change comes with the shared reading:
`extractTextFromChildren` trims, so whitespace-only children stop counting as
text. For the question `Button` and `Link` ask — is this a labelled control or
an icon-only one — that is the right answer, and a test pins it. The boolean no
longer short-circuits at the first text child, which does not matter at these
child counts.

`extractTextFromFirstChild`'s doc comment loses `Button`: #3053 replaced its
call with `containsTextChild`, so `Markdown`, `Initials` and `Truncate` are the
callers left that need the strict "exactly one text node" reading.

Separate commit (ba1b32ecf) on top of the merge, so the merge itself is a plain
union of both sides and the consolidation is reviewable on its own.

Still open: #3070 carries the same predicate as a local `isText` in
`Navigation.tsx`. Left untouched here — it is Lisa's PR, and
[the thread there](https://github.com/mittwald/flow/pull/3070#discussion_r3931406106)
has the two-line swap.

fixes #3028


packages/components/src/lib/react/remote.ts
```diff
@@ -16,6 +22,59 @@ export function isRemoteComponentRendererProps(
...
+
+      if (isRemoteTextRenderProps(child.props)) {
+        return child.props.remote.data;
+      }
+
+      return child.type === Fragment
+        ? joinTextChildren((child.props as PropsWithChildren).children)
+        : "";
+    })
+    .join("");
```
mfal, line 66: This walk is the third copy of the same classification, and it conflicts with the second.

- #3053 adds `isTextNode` (private) + an exported `containsTextChild` to this file — same branches plus fragment recursion, [now including `number`](https://github.com/mittwald/flow/pull/3053#discussion_r3931329843). It inserts at the same spot and makes the same React import change, so **this PR and #3053 conflict in `remote.ts`** and one of them rebases.
- #3070 carries a local `isText` in `Navigation.tsx`; [it drops it for `containsTextChild`](https://github.com/mittwald/flow/pull/3070#discussion_r3931406106).

Suggested order: #3053 → #3070 → this one, so nothing duplicated ever gets committed and this PR — the one with the most context on the topic — carries the consolidation.

On the rebase, `containsTextChild` collapses onto this walk and `isTextNode` disappears; one traversal covers both questions:

```ts
/** Whether any of `children` renders text. Fragments are transparent. */
export const containsTextChild = (children: ReactNode): boolean =>
  extractTextFromChildren(children) !== undefined;
```

`joinTextChildren` stays exactly as written here.

Two deltas worth deciding rather than inheriting silently:

1. **Whitespace.** `extractTextFromChildren` trims, so `containsTextChild(" ")` flips from `true` to `false`. For #3053's question — does this Button have a label — that reads correct to me: a whitespace-only child is not a label, and it should get the icon-only spacing. Worth a line in that PR's reasoning either way.
2. **No short-circuit.** The boolean answer now builds the string instead of stopping at the first hit. Irrelevant at the child counts these components see.

The `remotePackage` migration entry and the tests are unaffected by all of this.

packages/components/src/lib/react/remote.ts
```diff
@@ -16,6 +22,59 @@ export function isRemoteComponentRendererProps(
...
+
+/**
+ * The text of `children` when they are exactly one text node — locally a
+ * string, remotely a `RemoteTextRenderer` element. `undefined` for anything
+ * else, including a single string next to anything at all.
+ *
+ * That strictness is the point for its callers: `Button` decides its icon-only
+ * layout by it, `Markdown` takes it as the markdown source, and `Initials` and
+ * `Truncate` operate on the whole content. For "what does this read as", use
+ * {@link extractTextFromChildren}.
```
mfal: `Button` drops out of this list once #3053 merges — it replaces its `extractTextFromFirstChild` call with `containsTextChild`. `Option` is already gone via this PR, so `Markdown`, `Initials` and `Truncate` are what is left to justify the strictness.

Apply on the rebase, not before:

```suggestion
 * That strictness is the point for its callers: `Markdown` takes it as the
 * markdown source, and `Initials` and `Truncate` operate on the whole content.
 * For "what does this read as", use {@link extractTextFromChildren}.
 */
```

The argument itself holds unchanged — three callers still depend on the "exactly one text node" reading, which is why this stays a separate helper rather than a widened one.

packages/components/src/lib/react/remote.ts
mfal: Done in ba1b32ecf. `containsTextChild` is now `extractTextFromChildren(children) !== undefined` and `isTextNode` is gone; `joinTextChildren` stayed as written.

On the two deltas: the whitespace change is in, with a test — `containsTextChild("   ")` is `false`, which keeps a blank-labelled Button in its icon-only layout. Nothing in #3053's own browser tests depended on the old reading (components: 39 files / 268 pass). The lost short-circuit is noted in the commit body and left as is.

The merge commit (02bd80929) is a plain union of both sides, so the consolidation is reviewable on its own.

packages/components/src/lib/react/remote.ts
mfal: Applied in ba1b32ecf, together with the same correction in `remote.test.tsx`'s comment above the `extractTextFromFirstChild` block, which named `Button` too.

Lisa18289 (approved):

PR #3079, opened 2026-09-02
ci: run every suite on Dependabot PRs and deploy their review apps

mfal:
Two gaps on a Dependabot PR, both rooted in the same fact: `secrets.*` in a Dependabot-triggered run resolves against the repository's **Dependabot** secret store, not the Actions one.

## Review apps

The images were always built and pushed — only the deploy job was skipped. `cleanup-previews.yml` had it worse: it ran without a guard and failed on **every** Dependabot merge with `Missing required environment variables`, after the merge, where nobody looks.

Fixed in the two existing workflows, no Dependabot-specific ones:

- Only `MITTWALD_API_TOKEN` is a secret, and it is mirrored into the **Dependabot secret store**, since that is what `secrets.*` resolves against in those runs. The project id and the two TLS certificate ids are **repository variables** — not sensitive, and `vars.*` is one store for every run. Both are already configured.
- The deploy and cleanup jobs check out `pull_request.base.sha` when the actor is Dependabot. `pnpm install` then resolves the base lockfile, so the bumped package's code never executes in a job that holds those credentials. The build job is what exercises the PR's actual dependencies, and it holds none.
- The actor guards in both workflows are gone; the fork guard stays, since a fork gets no secrets from either store.

Scoped to Dependabot on purpose. On a human PR the deploy still runs the branch's own tooling, so a change to `dev/deploy-review.ts` is exercised by the PR that makes it. The same argument for pinning the deploy to a trusted revision applies to human PRs too — a reasonable follow-up, deliberately not in this PR.

## Tests

The full cross-version suite — both harnesses against every published version — was the one suite a Dependabot PR never ran (`skipping` on all four open ones). It was label-only, while visual has run unconditionally for a while. A dependency bump is exactly the change that moves rendered structure without touching a component.

- `test-cross-version-label.yml` now also runs on `opened`/`reopened`/`synchronize` for Dependabot PRs.
- `dependabot-auto-merge.yml` waits for **both** suites. Neither is a required status check in the ruleset, so this workflow is the only thing between a red run and a merge.

The gate reads each suite's verdict back for the PR head instead of trusting the payload, ordered by run `id` with `cancelled`/`skipped` dropped. That ordering is not pedantry: one commit carries four to five runs of the same suite, because Dependabot attaches its own labels right after opening the PR and every `labeled` event starts one that the concurrency group then cancels. Their `created_at` are equal to the second, and the surviving run is not necessarily the highest id.

## Corrected rationale

Several of these workflows claimed a Dependabot-triggered run gets a read-only `GITHUB_TOKEN`. The run log says `PullRequests: write`, and the ghcr push works — only the secrets are withheld. Comments fixed in four workflows (no behaviour change there), and CONTRIBUTE.md now records both halves of what such a run does and does not get.

## Configuration

Already in place, no action needed before merging:

- `MITTWALD_API_TOKEN` — Actions **and** Dependabot secret store.
- `MITTWALD_PROJECT_ID`, `MITTWALD_TLS_CERTIFICATE_ID_DOCS`, `MITTWALD_TLS_CERTIFICATE_ID_STORYBOOK` — repository variables.

The same ids still exist as Actions **secrets** from before this change. Nothing reads them any more, so they can be deleted whenever — deliberately not part of this PR.

Worth knowing for later: a *secret* the preview deploy needs has to be maintained in both stores, or Dependabot previews break while everything else stays green. A variable has no such trap, which is the reason to keep anything non-sensitive in `vars.*`. CONTRIBUTE.md records both halves.

#3125 moved these same ids the other way — `vars.*` → `secrets.*` — because the variables did not exist then, and the empty lookup had been silently falling back to ACME until the rate limit started serving placeholder certificates. This direction is only safe now that they are actually set; the warning #3125 added to `deploy-review.ts` stays as the net.

## Residual risk

A Dependabot PR's workflow *file* still comes from the PR branch, and this repo lets Dependabot bump GitHub Actions. A compromised action release is therefore a path to the deploy credentials that the base checkout does not close — the `cooldown: 7` in `dependabot.yml` is what stands in front of it. The previous `workflow_run` design closed that path, at the cost of two extra workflows that could not be tested from the PR introducing them. Deliberate trade-off: the preview apps are a disposable, containerised environment, which does not justify two workflows of machinery.

## Verifying this

Unlike the `workflow_run` design this replaces, the changed deploy path is live on this very PR — its own preview deploy exercises it (as a non-Dependabot run, so via the unchanged branch checkout). The Dependabot branch of the condition only becomes observable after merge: `@dependabot recreate` on one of the open bump PRs produces a build → deploy chain to watch.

## Cost

The cross-version suite is unsharded (~25 min, 45 min timeout) and now runs on every Dependabot PR, plus the cancelled label-storm runs. At four PRs a week with the rebase cascade, that is the most noticeable part of this change.


maaaathis (changes requested):
If I understand the security concern correctly, we want to make sure that the deployment credentials are never exposed to code or dependencies from the Dependabot PR.

But do we actually need separate deploy and cleanup workflows for that?

Couldn't the deploy job always check out the trusted base revision (e.g. `github.event.pull_request.base.sha`) before running `deploy-review.ts`, while the build job still checks out the PR and builds the actual preview images?

We could then provide `MITTWALD_API_TOKEN` and `MITTWALD_PROJECT_ID` under the same names as Dependabot secrets. This would keep the security boundary between PR code and deployment credentials while allowing us to reuse the existing deployment workflow instead of maintaining a separate Dependabot-specific one.

https://docs.github.com/en/code-security/reference/supply-chain-security/troubleshoot-dependabot/dependabot-on-actions#accessing-secrets

mfal:
You're right, and I've rewritten it that way — 302 lines gone, 51 added. Both Dependabot-specific workflows are deleted; the existing deploy and cleanup jobs now do the work, checking out `pull_request.base.sha` when the actor is Dependabot. Your read of the boundary holds: `pnpm install` resolves the base lockfile, so the bumped package never executes in a job that holds the credentials.

Two things that came out of working through it.

**It needs the secrets in both stores.** `MITTWALD_PROJECT_ID`, `MITTWALD_API_TOKEN` and the two `MITTWALD_TLS_CERTIFICATE_ID_*` have to be mirrored into the Dependabot store before this merges, or Dependabot previews fail on missing credentials. Noted in the PR description and in CONTRIBUTE.md, since it's a maintenance trap: adding a `MITTWALD_*` secret in only one store from now on breaks Dependabot previews while everything else stays green.

**One path stays open that the `workflow_run` version closed.** On a `pull_request` event the workflow *file* comes from the PR branch, and we let Dependabot bump GitHub Actions — so a compromised action release reaches a job that now holds the deploy credentials, and the base checkout doesn't help there (it pins the code, not the workflow). `tj-actions/changed-files` last March was exactly that shape.

I don't think that changes the decision. The preview apps are a disposable containerised environment, `cooldown: 7` sits in front of the realistic version of it, and it doesn't justify two workflows that can't even be exercised from the PR that introduces them. But it's the thing to remember if the preview project ever gets a credential that can reach further than it can today — worth a second look then.

I also kept the base checkout scoped to Dependabot rather than applying it to every PR. Unconditionally, a PR that changes `dev/deploy-review.ts` would no longer exercise its own change. Extending it is a sensible follow-up, just not silently in this one.

And a bug the rewrite surfaced on the way: the TLS certificate ids were being read from `vars.*` when they're stored as secrets — they'd have resolved empty and deployed the preview without its certificate, silently, since `deploy-review.ts` treats the id as optional.

maaaathis (changes requested):
One more thought: do `MITTWALD_PROJECT_ID` and the TLS certificate IDs actually need to be secrets?

As far as I can tell, only `MITTWALD_API_TOKEN` contains sensitive information. The project/certificate IDs can just be repository variables instead (`vars.*`).

I've already added the three IDs as repository variables and the API token as a Dependabot secret, so we'd only need to switch the workflow over to use `vars.*` for those three values.

mfal:
Agreed, and done in `ad19cb0` — `MITTWALD_PROJECT_ID` and the two certificate ids now read from `vars.*`, only the API token stays a secret. I confirmed the three variables and the Dependabot secret are in place before switching.

The nicer part is that it removes the trap rather than just tidying up: `vars.*` is one store for every run, so three of the four values can no longer be added in the Actions store alone and silently break Dependabot previews. CONTRIBUTE.md now says that explicitly, and that anything non-sensitive belongs in `vars.*` for exactly that reason.

One piece of history worth attaching to this, since it argues both ways: #3125 moved these same ids in the opposite direction, `vars.*` → `secrets.*`, because the variables didn't exist yet — the empty lookup had been silently falling back to ACME for a while, until enough previews had piled up to hit the rate limit and the hosts started serving the ingress controller's placeholder certificate. So the direction is right, it just needed the variables to actually exist first. They do now, and the warning #3125 added to `deploy-review.ts` stays as the net under it.

The three ids still exist as Actions secrets from before. Nothing reads them any more — deletable whenever, I've left them alone.

maaaathis (approved):

PR #3080, opened 2026-09-02
fix(ci): stop publishing on package-local non-shipping paths

### LINKED ISSUE #3023: Patch releases with an empty changelog: relevance is path-based, the changelog is type-based
**Status 2026-09-17 — both halves are merged and on `main`; this is ready to close.** #3098 took the path half, #3137 the type half. Everything below is struck through, with the verification noted at the end.

Four of the fourteen post-1.0 releases carry no changelog entry at all — `1.0.4`, `1.0.11`, `1.0.12` and `1.0.14` all read `**Note:** Version bump only for package @mittwald/flow-project`. There are two separate causes behind that, and only one of them is a bug.

## Why it happens

Two independent criteria decide, and they can never agree:

- **Whether a push releases is decided by PATHS.** The `decide` job in `publish.yml` classifies the changed files with `.github/scripts/release-relevance-lib.mjs`. Its denylist knows only repo-**root** prefixes (`.github/`, `apps/`, `docs/`, `dev/`, …), so everything under `packages/**` is publish-relevant wholesale. The commit type plays no role.
- **What lands in the changelog is decided by TYPE.** The `conventionalcommits` preset marks `docs`, `style`, `chore`, `refactor`, `test`, `build` and `ci` as `hidden: true`. Only `feat`, `fix`, `perf` and `revert` render. Nothing left to write → lerna emits "Version bump only".
- **The bump happens either way.** `lerna version --force-publish --conventional-commits`: the preset's `whatBump` returns level 2 (patch) for any non-empty commit range unless `bumpStrict` is set, and `--force-publish` bypasses changed-package detection by design (#2887).

## Evidence

| Version | Trigger | Reaches a tarball? |
| --- | --- | --- |
| 1.0.14 | `build(components)` #3010 — only `packages/components/.storybook/preview.tsx` | **No.** `files: ["*.md", "dist"]`; the Storybook config is never built |
| 1.0.12 | `ci:` #3006 — in `remote-react-components`: `package.json` (`scripts` only, `--project=unit` → `unit*`), `CONTRIBUTE.md`, `dev/cross-version/**`, `src/tests/visual/**` | **No.** `files: ["dist"]` |
| 1.0.11 | `docs(ColumnLayout)` #2954 — incl. `packages/components/AGENTS.md` | **Yes** — `*.md` ships. The release is correct by design; `docs:` is merely invisible |
| 1.0.4 | `docs: upgrade fumadocs-mdx` #2959 — an `apps/docs` dependency + the lockfile | No — **already fixed** by #2984 |

## Two problems, not one

1. **Needless releases** (1.0.14, 1.0.12). The relevance gate is too coarse *inside* `packages/**`. `.storybook/`, package-local `dev/`, `e2e/`, `src/tests/` and a `scripts`-only diff in a package manifest provably cannot affect a consumer. The `scripts` case stings twice: `classifyRootManifestChange` exists for exactly that reason (#2970) but is applied to the **root** manifest only.
2. **Real releases with an empty changelog** (1.0.11, and every `chore(deps):` bump). `chore(deps): bump framer-motion …` changes what consumers resolve — the changelog never says so. Suppressing those releases would be wrong; the changelog is what has to admit them.

The denylist direction stays as it is: unknown paths are relevant, so a forgotten docs path costs one needless version while a forgotten source path would swallow a real release. Only paths that are provably not consumer-visible get added.

## What to check

1. ~~Extend the denylist with package-local non-shipping paths: `packages/*/.storybook/**`, `packages/*/dev/**`, `packages/*/e2e/**`, `packages/*/src/tests/**`, plus `*.stories.tsx` and `*.test.*`. `IRRELEVANT_PREFIXES` is matched with `startsWith` today, so per-segment or glob matching is new machinery — keep it minimal and cover it with tests.~~ → #3098: `IRRELEVANT_PACKAGE_LOCAL_DIRS` (`.storybook`, `e2e`, `src/tests`, `dev/cross-version`, `dev/vitest`), `IRRELEVANT_PACKAGE_LOCAL_FILES` (`CONTRIBUTE.md`, `Dockerfile`, `.dockerignore`) and patterns for `*.stories.tsx` / `*.test.*`, matched segment-exact rather than with `startsWith`.
2. ~~Apply the key-diff refinement to `packages/*/package.json` too: a `scripts`-only change cannot reach a consumer. `classifyRootManifestChange` is reusable, but the `decide` job then has to fetch before/after for **every** changed package manifest, not just the root one. Note that published tarballs *do* contain `scripts` — the argument is "no consumer effect", not "not in the tarball". Write that down where the code says it.~~ → #3098: `classifyRootManifestChange` became `classifyManifestChange`, applied to the root manifest and every `packages/*/package.json`. The "no consumer effect, not 'not in the tarball'" reasoning is in the source comment.
3. ~~**Do not gate on commit type.** `chore(deps):` and `refactor:` routinely change the tarball; a type gate would silently swallow those releases. This is why #2931 chose paths.~~ → honoured: the gate stayed path-based, and #3137 solved the changelog half by *unhiding* types rather than gating on them.
4. ~~Decide what to do about problem 2. Either give `chore`/`build`/`refactor` a visible changelog section (`changelogPreset` in `lerna.json` accepts an object with `types`), or accept "Version bump only" as the honest signal for them and document it. Note the preset is pinned to `^9` for the reason in `docs/release-workflow.md`.~~ → #3137 took the first option: `lerna.json`'s `changelogPreset` is now an object respelling the full `types` list with every `hidden` dropped, so no type is invisible. `changelog-preset.test.mjs` asserts it.
5. ~~Verify the `next` line behaves symmetrically — it runs through the same `decide` job.~~ → symmetric by construction: `publish.yml` is one file for both lines (`on.push.branches: [main, next]`), because npm's Trusted Publisher binds one workflow filename per package.

## Acceptance criteria

- ~~A merge touching only `packages/*/.storybook/**`, `packages/*/dev/**`, `packages/*/e2e/**`, `packages/*/src/tests/**`, or carrying a `scripts`-only package-manifest diff, produces **no** npm publish, no tag, no GitHub Release and no `chore(release):` commit.~~
- ~~A merge touching package source, locales, SCSS, generated code, a package's shipped `*.md` or any dependency block publishes exactly as today.~~
- ~~A **mixed** merge publishes.~~
- ~~`workflow_dispatch` still forces a publish — the manual escape hatch stays.~~ (`publish.yml`: `::notice::Manual dispatch — publishing regardless of changed paths.`)
- ~~The skip is visible in the run (`::notice::`) and names the rule that fired.~~
- ~~`release-relevance-lib.test.mjs` covers every new path, incl. the mixed case and the `scripts`-only manifest.~~ (36 tests there, 4 in `changelog-preset.test.mjs`)
- ~~`docs/release-workflow.md` updated: the denylist section names the new package-local exclusions and the reasoning.~~ — plus a refinement this issue did not ask for: a package's root-level Markdown is judged against that package's own `files`, so `remote-react-components`' `AGENTS.md` (unpublished) is irrelevant while `flow-react-components`' is not.

## One case stays, by design

A release whose cause lies **outside every package** — `pnpm-lock.yaml`, the root manifest, `lerna.json` — still reads "Version bump only" (1.1.11, a lock-only Rollup bump). Lerna attributes commits per package, so with no package changed there is nothing to attribute. Documented in `docs/release-workflow.md` as the honest signal, not a bug.

## References

- #2931 — the original relevance gate
- #2984 — root manifest / lockfile refinement
- #2970 — the root `scripts` case that cut 1.0.9
- `docs/release-workflow.md` § "Not every merge releases"

mfal:
Release relevance is decided by PATH, the changelog by TYPE, and the two never agreed. This fixes both halves.

## Problem 1 — needless releases

The denylist knew only repo-ROOT prefixes, matched with `startsWith`, so everything under `packages/**` was publish-relevant wholesale. #3010 cut 1.0.14 from `packages/components/.storybook/preview.tsx` alone. #3006 cut 1.0.12 from a scripts-only package manifest, a package-local `CONTRIBUTE.md`, `dev/cross-version/**` and `src/tests/visual/**`. None of it reaches a consumer.

**Package-local denylist entries**, segment-exact under `packages/<name>/` — `e2e-helpers/` is not `e2e/`:

- `.storybook/**`, `e2e/**`, `src/tests/**`, `dev/cross-version/**`, `dev/vitest/**`
- the package's `CONTRIBUTE.md` (no package lists it in `files`)
- `*.stories.tsx` and `*.test.*` anywhere — `.test.` is not always the last extension (`*.browser.test.remote.tsx`)

`isPublishRelevant` is now a wrapper around `classifyPath`, which also returns the **rule** that decided. That is what the skip notice prints.

**The key-diff refinement now applies to every `packages/*/package.json`**, not just the root one — a `scripts`-only diff cannot reach a consumer. `classifyRootManifestChange` → `classifyManifestChange` (one extra `path` argument for the reason), and the `decide` job fetches before/after for each changed root/`packages/*` manifest into `MANIFEST_SNAPSHOT_DIR`. `apps/*` manifests are skipped — they are already irrelevant by path.

Published tarballs **do** contain `scripts`. The argument is "no consumer effect", not "not in the tarball" — nothing a consumer installs, imports or runs reads a package's `test:unit`. Same for the story and test `.d.ts` that `unplugin-dts` emits: they ship, and no `exports` path reaches them. Written into the code where the rules live.

### `packages/*/dev/**` is deliberately NOT denylisted

The issue proposed it; the repo says no. In `components` and `codemods` that directory **is** the build:

- `dev/vite/*` — the PostCSS/rollup plugins `vite.build.config.ts` imports; a change there moves the emitted CSS
- `dev/createDocPropertiesJson.ts` — writes the shipped `dist/assets/doc-properties.json`
- `dev/remote-components-generator/**` — generates `view.ts` and `src/auto-generated/**`
- `packages/codemods`' whole `build` script is `tsx dev/generateCli.ts && tsc … && tsx dev/buildTransforms.ts`

Two of those produce **dist-only** artifacts, so such a change arrives with no committed generated file beside it — a blanket `dev/**` entry would swallow that release silently, the one failure mode the denylist direction exists to prevent. Only the two provably test-only subtrees (`dev/cross-version/`, `dev/vitest/`) are listed, which is enough for the 1.0.12 evidence. Four counter-examples are pinned by tests.

Relevance stays **path-based**: no gate on commit type. `chore(deps):` and `refactor:` routinely change what consumers resolve.

## Problem 2 — real releases with an empty changelog

**Decision: unhide every type.** `lerna.json`'s `changelogPreset` becomes an object and respells the full `types` list with every `hidden: true` dropped.

Why not "accept Version bump only and document it": relevance is already decided by path, so after this PR a release exists **only** because something can reach a consumer. If a commit was worth a release it is worth a line. Suppressing those releases would be wrong — `chore(deps): bump framer-motion …` changes what consumers resolve, and a `docs:` commit on a shipped `AGENTS.md` really does change the tarball (#2954 → 1.0.11). Unhiding per type (`chore`/`build`/`refactor` only) would leave `test:`, `style:` and `docs:` able to release into an empty changelog; unhiding all leaves no residual case.

Cost is bounded: a section renders only when it has commits, and the types that previously released while hidden are now largely skipped by problem 1.

No bump semantics change: without `bumpStrict` the preset's `whatBump` returns patch for any non-empty commit range regardless of `hidden`. The preset stays pinned to `^9` per docs/release-workflow.md — `10.x` trips Lerna-Lite 5's legacy-writer path. Lerna-Lite passes the object straight to the preset factory (`GetChangelogConfig.getChangelogConfig` reads `changelogPreset.name` and forwards the object as `presetConfig`), so the auto-prefix to `conventional-changelog-conventionalcommits` still happens.

## Verification

- `node --test .github/scripts/*.test.mjs` — 63 pass; this is the command `test.yml` runs, and `publish.yml`'s `decide` job self-tests the classifier the same way. 17 new/changed relevance cases cover every new path, the mixed case, the scripts-only manifest at root and package level, and the segment-boundary negatives.
- One test reads every non-private `packages/*/package.json` and asserts `files` lists nothing but `dist` and shipped `.md`. That is the invariant the package-local denylist rests on; if a package starts shipping `src` or `CONTRIBUTE.md`, it fails.
- `changelog-preset.test.mjs` drives the real preset's `writer.transform` per type. Reverted to the old string preset it fails 4/4, so it is not vacuous.
- The guard was run end to end against fake snapshots for the 1.0.14 push, the 1.0.12 push, a mixed push, and a package manifest with no snapshot (fail-safe → publish).
- The `decide` job's `run` block: YAML parsed, `bash -n` clean, and the fetch loop exercised offline — slugs match the guard's `path.replaceAll("/", "__")`, `apps/docs/package.json` is not fetched, and a push with no manifest iterates zero times under `set -euo pipefail`.
- `pnpm lint` clean (0 errors; the 71 pre-existing `react-hooks/exhaustive-deps` warnings are untouched).
- `workflow_dispatch` is untouched: the `github.event_name != "push"` branch still returns `publish=true` before any classification.
- Both lines run through the same `decide` job, so `next` behaves identically — on `next` the classified set is what the `chore(sync):` forward-merge carried over. Symmetry verified by reading, not by a run.

Not verifiable locally: the actual `gh api` contents fetch and the rendered `::notice::` in a real run, and that a merge produces no tag/GitHub Release. Both need a push to a release line.

`SKIP_INSTALL_SIMPLE_GIT_HOOKS` already sits in `publish.yml`'s `env` and stays. No nx wiring needed — `.github/scripts/**` is outside the graph, and `test.yml` picks the new test up through its `*.test.mjs` glob.

fixes #3023

### DIFF
--- a/.github/scripts/changelog-preset.test.mjs
+++ b/.github/scripts/changelog-preset.test.mjs
@@ -1,0 +1,87 @@
+// @ts-check
+/**
+ * `lerna.json`'s `changelogPreset` renders EVERY commit type (#3023).
+ *
+ * Release relevance is decided by PATH, the changelog by TYPE, and the two can
+ * never agree: the `conventionalcommits` preset hides `docs`, `style`, `chore`,
+ * `refactor`, `test`, `build` and `ci` by default, so a release those types
+ * triggered had nothing to write and lerna emitted "Version bump only" — four
+ * of the first fourteen post-1.0 releases. `chore(deps):` bumps are the
+ * sharpest case: they change what consumers resolve and must stay releases, so
+ * the changelog is what has to admit them.
+ *
+ * The gate is `writer.transform`, which returns `undefined` for a hidden type.
+ * Asserting on it needs no git history and no rendering.
+ */
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import createPreset from "conventional-changelog-conventionalcommits";
+
+const { changelogPreset } = JSON.parse(
+  readFileSync(join(import.meta.dirname, "..", "..", "lerna.json"), "utf8"),
+);
+
+/** Every Conventional Commit type the repo's commit guard accepts. */
+const EXPECTED_SECTIONS = {
+  feat: "Features",
+  fix: "Bug Fixes",
+  perf: "Performance Improvements",
+  revert: "Reverts",
+  docs: "Documentation",
+  style: "Styles",
+  chore: "Miscellaneous Chores",
+  refactor: "Code Refactoring",
+  test: "Tests",
+  build: "Build System",
+  ci: "Continuous Integration",
+};
+
+test("changelogPreset: the preset is configured, not just named", () => {
+  // A bare string means the preset's defaults — which hide seven of the eleven
+  // types. The object form is what carries `types`.
+  assert.equal(typeof changelogPreset, "object");
+  assert.equal(changelogPreset.name, "conventionalcommits");
+  assert.ok(Array.isArray(changelogPreset.types));
+});
+
+test("changelogPreset: no configured type is hidden", () => {
+  for (const entry of changelogPreset.types) {
+    assert.notEqual(entry.hidden, true, entry.type);
+    assert.ok(entry.section, `${entry.type} has no section`);
+  }
+});
+
+test("changelogPreset: every commit type reaches the changelog", async () => {
+  const preset = await createPreset(changelogPreset);
+  const context = {
+    host: "https://github.com",
+    owner: "mittwald",
+    repository: "flow",
+  };
+
+  for (const [type, section] of Object.entries(EXPECTED_SECTIONS)) {
+    const commit = {
+      type,
+      scope: "components",
+      subject: "a subject",
+      hash: "0".repeat(40),
+      notes: [],
+      references: [],
+    };
+    const transformed = preset.writer.transform(commit, context);
+    assert.ok(transformed, `${type} is dropped from the changelog`);
+    assert.equal(transformed.type, section, type);
+  }
+});
+
+test("changelogPreset: a chore(deps) bump renders", () => {
+  // The case problem 2 of #3023 is about: it changes what consumers resolve,
+  // so it must publish AND say so.
+  assert.ok(
+    changelogPreset.types.some(
+      (entry) => entry.type === "chore" && entry.hidden !== true,
+    ),
+  );
+});
--- a/.github/scripts/release-relevance-guard.mjs
+++ b/.github/scripts/release-relevance-guard.mjs
@@ -6,19 +6,24 @@
  * Reads the changed-file list on stdin (one repository-relative path per line)
  * and writes `publish=true|false` to $GITHUB_OUTPUT.
  *
- * `ROOT_MANIFEST_BEFORE_FILE` / `ROOT_MANIFEST_AFTER_FILE` optionally point at
- * the two versions of the root `package.json`. They let the classifier judge
- * that manifest by its key diff instead of its path (#2970); without them the
- * root manifest stays relevant, as before.
+ * `MANIFEST_SNAPSHOT_DIR` optionally points at a directory holding both
+ * versions of every changed `package.json` the workflow fetched, as
+ * `<dir>/before/<slug>.json` and `<dir>/after/<slug>.json` with the path's `/`
+ * replaced by `__`. They let the classifier judge a manifest by its key diff
+ * instead of its path (#2970, #3023); a missing snapshot leaves that manifest
+ * relevant, as before.
  *
  * It never fails the run. The decision is a routing choice, not a policy
  * violation: an empty or unreadable input yields `publish=true`, which is the
  * behaviour before #2931.
  */
 import { appendFileSync, readFileSync } from "node:fs";
+import { join } from "node:path";
 import {
   classifyChangedFiles,
-  classifyRootManifestChange,
+  classifyManifestChange,
+  classifyPath,
+  isPublishRelevant,
 } from "./release-relevance-lib.mjs";
 
 /** Read all of stdin; an unreadable stdin is an empty list (fail-safe). */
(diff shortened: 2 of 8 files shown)


mfal:
Superseded by #3098 and closing — **this branch's commit is contained in it verbatim**, so nothing here is lost.

#3098 builds the next step on top: the PR title decides whether a merge releases and this classifier becomes its verifier, on the PR (a new blocking `release-intent` job) and again on the push (`decide` fails the release run on a mismatch instead of guessing). Rationale for the inversion: the type states intent, not effect — over the 80 merges since the gate landed, type and paths disagreed 21 times, and a type-only gate would have swallowed 8 real releases including the 1.1.0 promotion. Hence a `[release]` / `[no-release]` tag that only the contradicting titles carry.

It also adds `publishedDtsOptions` (story and test declarations out of `dist/types`, 196 of remote-react-components' 799 tarball entries) and closes #3083 — the 1.1.1/1.1.6 attribution question is answered there.

_🤖 Closed by [Claude Code](https://claude.com/claude-code)_

PR #3064, opened 2026-09-01
docs: split Upgrades off the Versionierung page

mfal:
## Summary
- Reverted the earlier rename — the page stays **Versionierung & Stabilität** / **Versionierung**.
- Split the upgrade-workflow content off into a new `/get-started/upgrades` page: the migration-guide pointer, the codemod CLI section, deprecation-warning guidance, and the post-release npm version-mismatch window. `Versionierung` keeps the semver contract itself (what's guaranteed, Node/React support, lifecycle status).
- Cross-linked the two pages, and updated the existing references in `installation`, `align`, and `segmented-control` to point at whichever of the two now actually covers that content.
- Links `Upgrades` from the `Releases` page's intro paragraph (both the rendered page and its shadow MDX used for search/`llms.txt`), so readers landing on release notes know where to find the upgrade how-to.

## Test plan
- [x] `pnpm nx test:links docs` — passes
- [x] `pnpm nx build docs` (full production build) — passes; verified `/get-started/upgrades` is generated, TypeScript compiles, and the split content lands on the right page in the built HTML
- [x] `prettier --check` on the touched files — passes

PR #3081, opened 2026-09-02
fix(ci): no release for merges that cannot reach a consumer (fixes #3023)

### LINKED ISSUE #3023: Patch releases with an empty changelog: relevance is path-based, the changelog is type-based

mfal:
Relevance was decided by repo-**root** paths, so everything under `packages/**` counted as publish-relevant wholesale. Two of the empty-changelog releases in #3023 came from that, and one class the issue does not list came later.

## What no longer publishes

Per PUSH, since that is what the `decide` job classifies:

| Push | What it carried | Version it cut |
| --- | --- | --- |
| #3010 | only `packages/components/.storybook/preview.tsx` | 1.0.14 — would not be cut now |
| #3008 | the Storybook image's `Dockerfile` / `.dockerignore` | folded into 1.0.15, which also carried #3013 (a real change) and would still be cut |

Two cases from #3023's evidence table are **not** closed, and that is deliberate — see "Not in scope":

| Push | Why it still publishes |
| --- | --- |
| #3006 (cut 1.0.12) | its `package.json` `scripts` diff and `src/tests/visual/**` drop out, but `remote-react-components/CONTRIBUTE.md` and `dev/cross-version/crossVersionRunner.ts` stay relevant |
| #3020 (cut 1.1.5) | `patches/` for `@mfalkenberg/react-live-ssr`, a dependency only `apps/docs` imports |

## The change

**The criterion is consumer effect, not tarball membership.** The two differ in both directions, and only the first decides: a tarball carries the package's `scripts`, yet nothing but the install lifecycle runs on a consumer's machine.

- **Package-local denylist:** `.storybook/**`, the Storybook image's `Dockerfile`/`.dockerignore`, `e2e/**`, `src/tests/**`, `*.stories.*`, `*.test.*`.
- **Every `package.json` is judged by its key diff**, not only the root one. `scripts` is irrelevant unless the diff touches `preinstall`, `install`, `postinstall` or `prepare`. No Flow package defines one today — which is exactly why the guard belongs in code and not in someone's memory. The `decide` job now fetches before/after for every changed manifest.
- **`packages/*/dev/**` stays relevant**, although #3023 proposed skipping it. It is a build input: `components/dev/vite/*` is imported by `vite.build.config.ts` and shapes the emitted CSS layers, `dev/component-index` and `dev/status-registry` write shipped `dist/assets/*.json`, and `codemods/dev/generate` writes the published `MIGRATION.md`.
- **A `stories/` directory stays relevant too** — only the story files are skipped. `createDocPropertiesJson.ts` parses every `.tsx` under `src/` and ignores only `*.stories.tsx`, so `components/Button/stories/lib.tsx` contributes five entries to the published `doc-properties.json`.
- **One gap stays open by choice:** a `patches/` entry for a dependency only a private project imports (1.1.5). Ruling it out needs the lockfile's importer graph including transitive edges, where a wrong answer swallows a real release. A test pins the current behaviour so the gap stays deliberate.

**Tests and stories leave `dist/types`.** `unplugin-dts` emitted declarations for all of `src`:

| 1.1.10 | tarball entries | of those stories/tests |
| --- | --- | --- |
| `flow-react-components` | 4160 | 126 `*.stories.d.ts` + 126 maps |
| `flow-remote-react-components` | 799 | 196 under `dist/types/tests/**` |

All eight release builds now share `publishedDtsOptions` from `packages/core`. **No consumer loses a type:** no package's `exports` has a wildcard subpath, so those declarations were unreachable. The globs cannot move into the shared tsconfig preset instead — its `exclude` governs `tsc --noEmit`, and that would stop type-checking the tests.

`@mittwald/flow-core` is now a declared devDependency of `ext-bridge`, `remote-core` and `remote-elements`, which newly import it. That is also the nx graph edge their build cache needs in order to invalidate on a `core` change.

## Verification

- Replayed **every** non-release commit on `main` since #2933 through the changed classifier: #3010 and #3008 flip to no-publish, and no commit that produced a legitimate release flips with them.
- `next` is symmetric — docs-only `chore(sync):` merges produce no `-next.N` bump (#3067, #3064, #3056, #3054 among others).
- 28 classifier tests, `build` and `test:compile` green for all eight touched packages; after the build, story/test declarations in `dist/types` drop from 252+6 to 18 in `components` and from 196 to 2 in `remote-react-components`. The remainder are real source modules the doc-properties generator sees, matching the denylist exactly.
- The `decide` job's shell loop was exercised locally against the no-match case (`set -euo pipefail` plus `grep || true`).

## Not in scope

Tracked in #3083:

- **The type-based half of #3023** — real releases whose changelog reads "Version bump only", because relevance is path-based while the changelog is type-based. Since #3023 was filed, 1.1.5, 1.1.7, 1.1.8 and 1.1.11 joined that list. Needs a `changelogPreset` decision, and an explanation for why 1.1.6 appears in the components changelog while 1.1.1 does not.
- **The relevance residue** — package-local non-shipped Markdown (judged against that package's `files`) and the test-only subdirectories of `dev/`. Those two are why #3006 and #3003 still publish.

Fixes #3023.

### DIFF
--- a/.github/scripts/release-relevance-guard.mjs
+++ b/.github/scripts/release-relevance-guard.mjs
@@ -6,10 +6,11 @@
  * Reads the changed-file list on stdin (one repository-relative path per line)
  * and writes `publish=true|false` to $GITHUB_OUTPUT.
  *
- * `ROOT_MANIFEST_BEFORE_FILE` / `ROOT_MANIFEST_AFTER_FILE` optionally point at
- * the two versions of the root `package.json`. They let the classifier judge
- * that manifest by its key diff instead of its path (#2970); without them the
- * root manifest stays relevant, as before.
+ * `MANIFEST_DIR` optionally points at a directory holding the two versions of
+ * every changed `package.json`, as `<path with / replaced by __>.before.json`
+ * and `.after.json`. They let the classifier judge each manifest by its key
+ * diff instead of its path (#2970, #3023); a manifest whose pair is missing
+ * stays relevant, as before.
  *
  * It never fails the run. The decision is a routing choice, not a policy
  * violation: an empty or unreadable input yields `publish=true`, which is the
@@ -18,6 +19,7 @@
 import { appendFileSync, readFileSync } from "node:fs";
 import {
   classifyChangedFiles,
+  classifyPackageManifestChange,
   classifyRootManifestChange,
 } from "./release-relevance-lib.mjs";
 
@@ -36,34 +38,43 @@ const files = readStdin()
   .filter((line) => line !== "");
 
 /**
- * The root manifest's two versions, if the workflow fetched them.
+ * One side of a manifest the workflow fetched, if it is there.
  *
- * A missing or unreadable file is `undefined`, which
- * `classifyRootManifestChange` reports as relevant — the behaviour before this
- * refinement.
+ * A missing or unreadable file is `undefined`, which the classifiers report as
+ * relevant — the behaviour before this refinement.
  *
- * @param {string} variable
+ * @param {string} path
  */
-function readManifest(variable) {
-  const path = process.env[variable];
-  if (!path) return undefined;
+function readManifest(path) {
   try {
     return readFileSync(path, "utf8");
   } catch {
-    console.log(`::warning::Could not read ${variable} (${path}).`);
     return undefined;
   }
 }
 
-/** @type {{ rootManifestRelevant?: boolean }} */
-const options = {};
-if (files.includes("package.json")) {
-  const manifest = classifyRootManifestChange(
-    readManifest("ROOT_MANIFEST_BEFORE_FILE"),
-    readManifest("ROOT_MANIFEST_AFTER_FILE"),
-  );
-  options.rootManifestRelevant = manifest.relevant;
-  console.log(`Root manifest: ${manifest.reason}.`);
+/** @type {{ manifestRelevance: Record<string, boolean> }} */
+const options = { manifestRelevance: {} };
+const manifestDir = process.env.MANIFEST_DIR;
+
+for (const path of files.filter(
+  (file) => file === "package.json" || file.endsWith("/package.json"),
+)) {
+  const slug = path.replaceAll("/", "__");
+  const before = manifestDir
+    ? readManifest(`${manifestDir}/${slug}.before.json`)
+    : undefined;
+  const after = manifestDir
+    ? readManifest(`${manifestDir}/${slug}.after.json`)
+    : undefined;
+
+  const { relevant, reason } =
+    path === "package.json"
+      ? classifyRootManifestChange(before, after)
+      : classifyPackageManifestChange(before, after, path);
+
+  options.manifestRelevance[path] = relevant;
+  console.log(`Manifest: ${reason}.`);
 }
 
 const { publish, reason, relevant, total } = classifyChangedFiles(
--- a/.github/scripts/release-relevance-lib.mjs
+++ b/.github/scripts/release-relevance-lib.mjs
@@ -2,10 +2,20 @@
 /**
  * Release-relevance classification — pure functions, no git / no IO.
  *
- * Answers one question: can this set of changed files reach a PUBLISHED
- * package's tarball? Docs- and CI-only merges cannot, and must not produce a
- * release — no npm publish, no `chore(release):` bump, no tag, no GitHub
- * Release (#2931).
+ * Answers one question: can this set of changed files change what a CONSUMER of
+ * a published package gets? Docs- and CI-only merges cannot, and must not
+ * produce a release — no npm publish, no `chore(release):` bump, no tag, no
+ * GitHub Release (#2931).
+ *
+ * "Consumer effect", not "inside the tarball" — the two are not the same, and
+ * the difference decides two whole classes (#3023). A published tarball carries
+ * the package's `scripts`, yet only the install lifecycle ones ever run on a
+ * consumer's machine. And `dist/types` is emitted from all of `src`, which used
+ * to drag test and story declarations along — 196 of `flow-remote-react-`
+ * `components@1.1.10`'s 799 entries were `dist/types/tests/**`, dead files no
+ * `exports` path can reach. The build excludes them now, but the criterion has
+ * to be consumer effect either way: nothing stops the next generator from
+ * emitting something similar.
  *
  * The rule is a DENYLIST, and that direction is deliberate. Every published
  * package ships `dist` (plus three `.md` for flow-react-components), so "what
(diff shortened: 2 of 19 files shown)


mfal:
Superseded by #3098 and closing.

#3080 had already taken the same ground further: it carries the package-local denylist including `CONTRIBUTE.md`, `dev/cross-version/**` and `dev/vitest/**` (which I had deferred to #3083), returns the rule that decided each skip, and fixes the changelog half through `changelogPreset`. It also sits on newer `main`.

Nothing is lost: #3098 contains #3080's commit and carries over the one thing this branch had that it lacked — the `publishedDtsOptions` cleanup that keeps story and test declarations out of `dist/types` — plus the lockfile correction from this branch's second commit.

_🤖 Closed by [Claude Code](https://claude.com/claude-code)_

PR #3097, opened 2026-09-02
feat(List): let the item and row text be selected and copied

### LINKED ISSUE #895: Copy content of the list
It should be possible to copy the content of the list by selecting the text. This is currently not possible because the [GridList](https://react-spectrum.adobe.com/react-aria/GridList.html) is used by Aria.

Lisa18289:
A ListItem and a table row are click targets, and React Aria sets
`user-select: none` on them while pressing, so their text could never be
selected. The item's title and subtitle and the table's cells now opt back in.

Dragging never activates an item, whether or not it selected text — React Aria
triggers the press from the click event, so selecting a heading would otherwise
activate the item, and a drag that missed the text would navigate away rather
than do nothing. An ordinary click still activates it, including one that
wobbles a few pixels or lands inside an existing selection. Drags starting on an
interactive child, and clicks arriving without a pointer interaction (keyboard,
screen reader), are left alone.

A linked item covers itself with an empty anchor so a right-click reaches a real
link, and that anchor sat over the title too — every press there landed on it,
so no selection could start. The title and subtitle now sit above the overlay,
the way interactive content and the bottom slot already did. The row keeps
owning activation, so clicking them follows the link as before; what moves is
the browser's own link menu, which now answers on the rest of the item rather
than on the text itself.

The title box hugs its text exactly, so a drag would have had to hit the text
pixel for pixel. The header around it is selectable as well, and the title
carries padding cancelled out by the same negative margin, so a drag can start
beside the text without moving it. Verified against Storybook: six stories,
including the long-word edge case, render byte-identical.

WebKit has no unprefixed `user-select`, so the rule only applies in Safari with
the autoprefixer from #3066. That is merged and part of this branch now.

The drag tests live in a new `browser-mouse` vitest project. Only a native drag
creates a real text selection, and parallel test files share one page and one
Playwright cursor, so another file's click releases the held button mid-drag.

The cursor over the text stays `pointer`. Making it `text` would weaken the
"this whole row is clickable" affordance, so that is a UX call rather than part
of this change.

Closes #895


Lisa18289:
Closing this: the approach turned out too hacky. We're considering a solution with copy buttons instead.

PR #3098, opened 2026-09-02
fix(ci): stop publishing on paths that cannot reach a consumer

mfal:
The release denylist knew only repo-**root** prefixes, matched with `startsWith`, so everything under `packages/**` counted as publish-relevant wholesale:

| Version | Trigger | Reaches a consumer? |
| --- | --- | --- |
| 1.0.14 | #3010 — `packages/components/.storybook/preview.tsx` alone | No |
| 1.0.12 | #3006 — a `scripts`-only package manifest, a package-local `CONTRIBUTE.md`, `dev/cross-version/**`, `src/tests/visual/**` | No |

**Package-local denylist entries**, segment-exact under `packages/<name>/` — `e2e-helpers/` is not `e2e/`:

- `.storybook/**`, `e2e/**`, `src/tests/**`, `dev/cross-version/**`, `dev/vitest/**`
- the package's `CONTRIBUTE.md` (no package lists it in `files`)
- `Dockerfile` and `.dockerignore` — the Storybook preview image, built by CI from `dev/` and `.github/`. #3008 pushed both and nothing else under `packages/`
- `*.stories.tsx` and `*.test.*` anywhere — `.test.` is not always the last extension (`*.browser.test.remote.tsx`)

A package's **root-level Markdown is judged against that package's `files`**, not by path — nothing builds one, so it reaches a consumer exactly if it is published. `flow-react-components` lists `AGENTS.md`, `MIGRATION.md` and `USAGE.md`; `remote-react-components` lists only `USAGE.md`, so ITS `AGENTS.md` reaches nobody. Markdown *deeper* in a package keeps its path-level relevance: `codemods/src/migrations/*/entry.md` is a generator input, not documentation.

`packages/*/dev/**` is deliberately **not** denylisted wholesale: in `components` and `codemods` that directory *is* the build — `dev/vite/*` holds the plugins `vite.build.config.ts` imports, `dev/createDocPropertiesJson.ts` writes the shipped `dist/assets/doc-properties.json`, `dev/remote-components-generator/**` generates `view.ts` and `src/auto-generated/**`, and `codemods`' build script is `tsx dev/generateCli.ts && …`.

The criterion is **no consumer effect**, not "not in the tarball": a tarball carries the package's `scripts`, yet nothing a consumer installs reads them.

**The key-diff refinement applies to every `packages/*/package.json`**, not just the root one — a `scripts`-only diff cannot reach a consumer. `publish.yml` therefore snapshots each changed manifest into `MANIFEST_SNAPSHOT_DIR` instead of the two `ROOT_MANIFEST_*` files. A failed fetch leaves the snapshot missing, which the classifier still reads as "relevant" — the behaviour before, and what an added or deleted manifest gets.

`isPublishRelevant` is now a wrapper around `classifyPath`, which also returns the **rule** that decided — that is what the skip notice prints.

## Verified

- 35 classifier tests, run by the `decide` job before it classifies.
- Replayed every non-release commit on `main` since #2933 through the changed classifier: #3010 and #3008 flip to no-publish, and no commit that produced a legitimate release flips with them.
- `next` stays symmetric — docs-only `chore(sync):` merges produce no `-next.N` bump.

`docs/release-workflow.md` is cut back to the mechanics: what the classifier excludes and what it judges by content.

## Split out of this PR

This PR was three changes; each now has its own line, all rebased onto current `main`.

- **#3103** (merged) — keeping tests and stories out of `dist/types`. Changes what the packages publish, not when.
- **the changelog half** — `changelogPreset` rendering every commit type. Relevance is decided by path, the changelog by type; two axes, two PRs.
- **the commit-mixture guard** — reviewed as #3102 and merged into *this branch* rather than into `main`. It is now standalone, and it **depends on this PR**: without the package-Markdown and `Dockerfile` rules above it would fire on #3009 and #3008, which reach nobody, and its `MANIFEST_SNAPSHOT_DIR` wiring needs the classifier below. Merge this one first.

A title-driven variant of the release gate (`[release]` / `[no-release]` tags in the PR title) was built and withdrawn — the tag puts the decision in a hand-written token while the paths answer the question directly. The branch name still carries that withdrawn approach; the PR number is kept for its review history.

Part of #3023 — the path half.

### DIFF
--- a/.github/scripts/release-relevance-guard.mjs
+++ b/.github/scripts/release-relevance-guard.mjs
@@ -6,19 +6,26 @@
  * Reads the changed-file list on stdin (one repository-relative path per line)
  * and writes `publish=true|false` to $GITHUB_OUTPUT.
  *
- * `ROOT_MANIFEST_BEFORE_FILE` / `ROOT_MANIFEST_AFTER_FILE` optionally point at
- * the two versions of the root `package.json`. They let the classifier judge
- * that manifest by its key diff instead of its path (#2970); without them the
- * root manifest stays relevant, as before.
+ * `MANIFEST_SNAPSHOT_DIR` optionally points at a directory holding both
+ * versions of every changed `package.json` the workflow fetched, as
+ * `<dir>/before/<slug>.json` and `<dir>/after/<slug>.json` with the path's `/`
+ * replaced by `__`. They let the classifier judge a manifest by its key diff
+ * instead of its path (#2970, #3023); a missing snapshot leaves that manifest
+ * relevant, as before.
  *
  * It never fails the run. The decision is a routing choice, not a policy
  * violation: an empty or unreadable input yields `publish=true`, which is the
  * behaviour before #2931.
  */
 import { appendFileSync, readFileSync } from "node:fs";
+import { join } from "node:path";
 import {
   classifyChangedFiles,
-  classifyRootManifestChange,
+  classifyManifestChange,
+  packageRootMarkdown,
+  shipsRootMarkdown,
+  classifyPath,
+  isPublishRelevant,
 } from "./release-relevance-lib.mjs";
 
 /** Read all of stdin; an unreadable stdin is an empty list (fail-safe). */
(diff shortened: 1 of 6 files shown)


mfal:

ins0 (approved):

Issue #1250, opened 2025-03-10
Linked ListItems can not be opened in a new tab


Lisa18289:
When right-clicking on a list item, the browser does not show the option to open the underlying link in a new tab. The link can only be accessed via the context menu. However, using `Cmd + Click `on the list item does open the link in a new tab — at least in Chrome.

https://github.com/user-attachments/assets/10d45bcb-6f3c-4d7f-8f1e-451233f23e10

PR #3111, opened 2026-09-03
fix(List): restore hover carve-out for list item bottom content

mfal:
## What & why

Hovering a list item's bottom content highlighted the whole item. The carve-out that should prevent this has been dead since the `View` → `ListItemView` move.

`Item.module.scss` matched `:global(.flow--list--items--item--view--bottom-content)`. That name was correct when the rule landed (#1250) — the bottom content was rendered by `Item/components/View` back then. Moving that render to `ListItemView` changed the generated class to `flow--list--list-item-view--bottom-content`, so the `:not(:has(…))` became vacuously true and the item highlighted on bottom-content hover again. Nothing flagged it: no type covers a `:global()` string, and `testScreenshot` resets the pointer position, so no visual baseline can capture a pure `:hover` style.

Rather than hard-coding the new generated name — which is exactly how this broke — `Item.module.scss` now owns the hook class and `ListItemViewContent` applies it alongside its own:

- `Item.module.scss`: `&:where(.hasAction):not(:has(.bottomContent:hover))` — a local class, no `:global()`.
- `ListItemViewContent.tsx`: the bottom slot gets `clsx(styles.bottomContent, itemStyles.bottomContent)`.

A rename now breaks `tsc` through the generated `Item.module.d.scss.ts`, and a path move cannot break it at all. Also drops the two dead `className` entries in `contentProps` that duplicated `getStyleForContentSlot`.

**The same hard-coded name appeared a second time** (spotted in review): the z-index lift that keeps the bottom slot clickable above the link overlay still selected `:global(.flow--list--list-item-view--bottom-content)`. Since `ListItemViewContent` now puts Item's own class on that exact element, that selector uses `.bottomContent` too. Same element, same specificity — behaviour-neutral — and `Item.module.d.scss.ts` no longer declares a foreign class name at all.

## Notes for reviewers

- The new browser test in `List.browser.test.tsx` asserts the item's computed `backgroundColor` across both hover targets, comparing against a **second, never-hovered item** rather than a "before" snapshot — the pointer may already rest on the item at render time, so its own initial state is not a reliable default-state baseline. It was written first and fails against the pre-fix code (`expected 'rgb(229, 239, 248)' to be 'rgba(0, 0, 0, 0)'`); it also fails if `itemStyles.bottomContent` is dropped from the className, so it guards both halves of the coupling.
- **No snapshot update needed.** The visual `List` suite passes unchanged (18/18, webkit) — no baseline covers a hovered bottom content. The `run-visual-tests` label is set for verification only.
- **Nothing consumers may target changes.** `flow--list--list-item-view--bottom-content` is still applied to the bottom slot — only Flow's own selectors stopped naming it from the outside. The added `flow--list--items--item--bottom-content` is additive.
- `Item.module.d.scss.ts` is generated (`pnpm nx build:scss-types components`); its diff is generator output, not a hand edit.

## Verification

Full suite, on the first two commits:

- `vitest --project=browser*` (components, webkit): 38 files / 263 tests pass
- `pnpm nx test:unit components`: 253 pass — `test:compile`: clean — `pnpm lint`: 0 errors
- `pnpm build`: no generated-code drift
- visual `List.browser.test.tsx` (remote-react-components, webkit): 18/18 against existing baselines

On the review follow-up (`612dffc`), re-run scoped to what it touches:

- `pnpm nx test:compile components`: clean
- `pnpm nx test:browser components --browser.name=webkit -- src/components/List/List.browser.test.tsx`: 64 pass, incl. the hover carve-out test
- `prettier --check` clean; `stylelint` reports only the pre-existing `selector-max-type` warning from #3132

## Checklist

- [x] `pnpm lint` is clean and `pnpm affected:test` passes (browser tests if behavior changed)
- [x] **Generated code is committed** (`git diff` is empty after the relevant `build:*` targets)
- [x] User-facing strings added to **both** `de-DE` and `en-US` locale files — n/a, no UI text
- [x] Docs updated if a public API changed; intentional visual changes get updated snapshots / the `update-screenshots` label — no public API change, no baseline change


packages/components/src/components/List/components/Items/components/Item/Item.module.d.scss.ts
```diff
@@ -2,7 +2,7 @@ declare const classNames: {
   readonly item: "item";
   readonly isSelected: "isSelected";
   readonly hasAction: "hasAction";
-  readonly "flow--list--items--item--view--bottom-content": "flow--list--items--item--view--bottom-content";
+  readonly bottomContent: "bottomContent";
   readonly "flow--avatar": "flow--avatar";
   readonly link: "link";
   readonly "flow--list--list-item-view--bottom-content": "flow--list--list-item-view--bottom-content";
```
Lisa18289, line 8: wird die Zeile noch benötigt? funktioniert das nicht auch mit bottomContent?

mfal: Guter Punkt — ja, das geht mit `bottomContent`: `ListItemViewContent` setzt Items eigene `bottomContent`-Klasse auf genau dasselbe Element wie ListItemViews. Der `:global(.flow--list--list-item-view--bottom-content)`-Selektor für den z-index-Lift nutzt jetzt `.bottomContent`, damit ist die Zeile aus der generierten Datei raus (612dffc).

Lisa18289 (approved):

PR #3136, opened 2026-09-08
feat(Heading): render a Button in the heading content

mfal:
## What & why

Put a `Button` or `CopyButton` **inside** a `Heading` and it now renders in the
heading content, right behind the headline:

```tsx
<List.ItemView>
  <Avatar><IconEmail /></Avatar>
  <Heading>
    {mail}
    <CopyButton text={mail} />
  </Heading>
  <Text>Rebel Alliance</Text>
  <ContextMenu>…</ContextMenu>
</List.ItemView>
```

Two entries in `Heading`'s props context: `Button` and `CopyButton` join
`Badge`, `AlertBadge` and `ContextualHelpTrigger` in the `headingContent`
tunnel, with an overridable `size` default: `s`, and `m` on a heading that
renders at `xl` or `xxl`.

`CopyButton` needs its own entry rather than riding along on `Button`'s: it is a
UI flow component, so its `ClearPropsContext` shields the `Button` it renders
internally from Heading's props context, and without the entry it stayed in the
heading text. Tunneling the `CopyButton` itself — not its inner button — keeps
its `TooltipTrigger` and `Action` wrapper together with it; clicking it in the
tunneled position still produces the Action's success feedback.

**Why it was impossible before.** `ListItemView` tunnels a `Heading` into the
title slot and a `Button` into the action slot at the far right, so the two
never end up next to each other. A `Combine` around them is torn apart by
exactly those tunnels — reproduced in Storybook before touching anything: the
`Combine` renders an empty div and the button flies to the right edge.

**Why not a `Combine` recipe.** `.headingContent` already solves the
arrangement — `inline-flex`, centered on the heading's line, wraps together with
the text, gap from `--heading--icon-to-text-spacing`. No new CSS, no new design
token. A `Combine` recipe would have needed a new `:has()` rule, a new spacing
token, **and** a conditional tunnel redirect in `ListItemView` (only redirect a
`Combine` that actually contains a `Heading`) — a second, competing way to the
same result.

The `s` default keeps the row the height of a plain heading row; a size-`m`
button visibly inflates it. A local `size` prop still wins over the context.

At `xl` and `xxl` that flips: there the small button reads as an afterthought
next to the text, so it scales to `m`. The branch runs on the **effective** size
— the `size` prop when given, otherwise the level's own font size through
`sizeByLevel` — so `level={1}` and `level={1} size="xl"` produce the same button
(review feedback; the first version keyed on the prop alone and gave two
visually identical headings differently sized actions). `sizeByLevel` is the
TypeScript half of the level rules in `Heading.module.scss` and has to be kept in
step with them.

Not reachable from a prop, and therefore still open: the `<=768px` media query
scales `size-xl` down to `l`, where the button stays `m`.

### Reviewer notes

- **Rendering change for consumers** who already nested a `Button` in a
  `Heading`: it moves from the text flow into the heading content. Nothing
  documented that position before, and `ContextualHelpTrigger`'s inner button is
  unaffected (it resets the tunnel and already forces `size: "s"`). Worth a line
  in the release notes.
- **The `s` default leaked into injected chrome** and CI caught it. `Accordion`
  and `Modal` inject their own button into a `Heading`'s children through a
  `children: dynamic` props-context entry, so the default shrank Accordion's
  chevron and Modal's close icon (and with it the modal's height). Only the
  size leaked, never the tunnel — both buttons already opt out. Accordion resets
  it through the props context (`Heading: { Button: { size: "m" } }`), because a
  local prop on its shared `HeaderButton` would also change the `Label` path,
  where that button renders at size `s` today. Modal isolates its close button
  with `ClearPropsContext` **plus** `ClearPropsContextView`, so a consumer's
  button in a modal heading still gets the `s` default — with only the host
  variant, Local went green while Remote still rendered the small icon.
- In the **tile** view the button lands top-right over the avatar image — the
  position `headingContent` already gives badges there. Consistent, but a
  deliberate consequence.
- **`AnchorLinkHeading`** in `apps/docs` laid its copy-link button out with
  `display: flex` on `heading-text`. With the button now in `headingContent`,
  that rule made the heading text block-level and pushed the button onto a
  second line. Rule dropped; hover reveal, focus reveal and the 8px gap verified
  unchanged (`--size-rem--s` and `--heading--icon-to-text-spacing` are both 8px).
- No generated code changed — this is a props-context entry, not a prop.

### Verified

Storybook (list view, 375px wrapping, tile view, size override) plus two new
visual scenarios that pass in **Local *and* Remote** and in **light *and* dark**:
both environments produce pixel-identical output against the same baseline, so
the tunnel survives serialization.

The **full visual suite** runs green locally in both browsers — 84 files, 368
tests each in webkit (light) and firefox (dark) — against the unchanged darwin
baselines, which is what establishes that nothing beyond the two new scenarios
renders differently. `components`' browser suite: 570 tests green.

The `xl`/`xxl` scaling moves exactly one scenario: `Heading with button`, 454
pixels (ratio 0.01), Local and Remote alike. Confirmed in the actual screenshot
that only the `xl` and `xxl` rows change — the smaller sizes and the icon-button
row are untouched. Deriving the size from the level adds no churn on top: the
scenario's only level-driven heading is a level 2, which resolves to `m` and
keeps its `s` button. `test:compile` green for both packages, no generated code
changed.

**Note for anyone re-running the visual suite by hand:** it resolves the **built**
`components` package, so `vitest run …` straight after a source edit tests the
previous build. The restored `CopyButton` first came out untunneled for exactly
that reason — `pnpm nx build components` (or the nx target) first.

## Base branch & title

`feat(Heading):` → base **`next`**, per the routing guard.

## Checklist

- [x] `pnpm lint` is clean; `pnpm nx test:unit components` and `test:compile`
      for `components` + `remote-react-components` pass; the two new visual
      scenarios pass on webkit and firefox
- [x] **Generated code is committed** (`git diff` empty after the generators)
- [ ] User-facing strings added to **both** `de-DE` and `en-US` locale files —
      n/a, no new UI text
- [x] Docs updated (`Heading` → "Kombiniere mit … Button", `ListItemView` →
      "Mit Button an der Heading", plus both examples)

> **Fourth `update-screenshots` round** for the restored `CopyButton` row.
>
> **Third round**, for the `xl`/`xxl` button scaling: the label run committed the
> two Linux baselines and nothing else (`3ee1268`) — reviewed the commit's file
> list. The two macOS ones it leaves stale are regenerated in `a72d652`, with the
> file filter ahead of `--update` so the flag cannot swallow it.

> **Second round** for the `Heading with button` Linux
> baselines, which the `CopyButton` row added to that scenario. This time the
> full suite is green locally in both browsers first, so the label bakes in
> nothing broken.

> **First round:** the `update-screenshots` run happened while the
> size leak above was still present, so it committed the Accordion, Modal and
> IllustratedMessage regressions as the new Linux truth (`--update` over the
> whole suite plus `git add -A`). Those 20 baselines are restored to `next`'s
> content in `4d5ca94`; the four genuinely new ones from that run are kept and
> were reviewed.


packages/components/src/components/Heading/Heading.tsx
```diff
@@ -64,6 +66,13 @@ export const Heading = flowComponent("Heading", (props) => {
         component: "Heading",
       },
     },
+    Button: {
+      tunnel: {
+        id: "headingContent",
+        component: "Heading",
+      },
+      size: buttonSize,
+    },
```
maaaathis, line 88: `CopyButton` seems to be broken with the current implementation. The PR description still states that both `Button` and `CopyButton` are supported, but `Heading` currently only provides props for `Button`. Since `CopyButton` is a UI flow component, its implementation is wrapped in a cleared props context, so the inner `Button` won't receive the `headingContent` tunnel.

The version before the force-push had an explicit `CopyButton` entry and visual coverage for this. Should we restore that using `buttonSize`, or is `CopyButton` no longer supposed to be part of this change?

packages/components/src/components/Heading/Heading.tsx
```diff
@@ -47,6 +47,8 @@ export const Heading = flowComponent("Heading", (props) => {
 
   const Element = elementType ?? Aria.Heading;
 
+  const buttonSize = size === "xl" || size === "xxl" ? "m" : "s";
```
maaaathis: Should the button size follow the rendered heading size rather than only the explicit `size` prop?

For example, `Heading level={1}` renders as `xl` through CSS but still gets an `s` button, while the visually equivalent `Heading level={1} size="xl"` gets an `m` button. That means two headings with the same visual size can produce differently sized actions.

maaaathis (changes requested):

packages/components/src/components/Heading/Heading.tsx
mfal: Restored in 8f2725a0 — the `CopyButton` entry is back in Heading’s props context (from 647def423 on the dropped round), now with `size: buttonSize` instead of the fixed `s`, plus the stories and the `Heading with button` coverage. It was lost in the branch rewrite, not deliberately dropped.

packages/components/src/components/Heading/Heading.tsx
mfal: Changed in bb40e43e — the size now follows the effective heading size: the `size` prop when given, otherwise the level’s own font size via `sizeByLevel`, so `level={1}` and `size="xl"` agree. Caveat in the commit: the `<=768px` media query scales `size-xl` down to `l` and the button stays `m`; that one is not reachable from a prop.

maaaathis (approved):

PR #3150, opened 2026-09-11
ci(publish): assert every package reached npm before calling a release done

ins0:
## What & why

A green publish run is not proof that a release reached npm. `lerna publish`
maps npm's `409 Failed to save packument` — the answer to a **concurrent write
on the same package document** — to "already published": it warns and exits 0.

That is how **1.1.31 shipped 10 of 13 packages** ([run 34473093304][run]) with
the tag and the GitHub Release cut, and nobody noticed for a day:

```
lerna-lite WARN publish Package is already published: @mittwald/flow-stylesheet@1.1.31
lerna-lite WARN publish Package is already published: @mittwald/flow-remote-elements@1.1.31
lerna-lite WARN publish Package is already published: @mittwald/flow-remote-react-renderer@1.1.31
lerna-lite success published 10 packages
```

None of the three was published. `flow-remote-react-components@1.1.31` — the
`latest` an extension developer installs — pinned
`flow-remote-elements@1.1.31` (`workspace:*` packs as an **exact** pin), so:

```
npm error code ETARGET
npm error notarget No matching version found for @mittwald/flow-remote-elements@1.1.31.
```

The three were backfilled by a dispatch with `only_publish` ([run
34577464400][fix]); 1.1.31 is complete again.

**The collision needs no bad luck.** `main` and `next` publish the *same* npm
packages; their concurrency groups (`mutate-main` / `mutate-next`) serialize
each line against itself but not against the other, and `forward-merge.yml`
starts the `next` run seconds after the `main` one. Both then walk the same
toposort with `--concurrency 1` at the same pace. The two runs drifted into
lockstep and collided on three packages, to the second:

| Time (UTC) | `next` → 1.2.0-next.37 | `main` → 1.1.31 |
| --- | --- | --- |
| 11:51:05 | ✅ `flow-stylesheet` | ⚠️ "already published" |
| 11:51:12 | ✅ `flow-remote-elements` | ⚠️ "already published" |
| 11:51:19 | ✅ `flow-remote-react-renderer` | ⚠️ "already published" |

## What this adds

`publish.yml` now asserts, as its **last** step, that every publishable package
resolves on the registry at its manifest version — and fails the run naming the
packages and the remedy when one does not.

- `publish-completeness-lib.mjs` — pure classification (package selection
  mirroring lerna's rule, lockstep check against `lerna.json`, presence
  classification, retry budget, verdict, report)
- `publish-completeness-guard.mjs` — fs/network shell; writes a per-package
  table to the job summary
- `publish-completeness-lib.test.mjs` — 12 `node:test` cases, run by `test.yml`
  and self-tested in the publish job before the publish
- `eslint.config.js` — `fetch` / `AbortSignal` added to the globals for
  `.github/scripts/**/*.mjs`

Three decisions worth recording:

**It runs last, after push/tag/release.** The remedy is a dispatch with
`only_publish`, which publishes whatever the branch's manifests say. Failing
before the push would leave the branch at N-1 while npm carries N — and the
remedy would then republish N-1, healing nothing.

**A 404 is a hole; anything else is not.** A 5xx, a rate limit or a timeout says
nothing about the version. It fails the run too, but as a distinct outcome whose
remedy is "re-run this job", never "republish".

**The retry budget is ~305s, not seconds.** Measured on the backfill above:
`flow-stylesheet` kept answering 404 for minutes after its own publish, while
the two packages published beside it in the same command were readable within
seconds. A guard that fires false reds gets ignored.

## Verified

- `node --test .github/scripts/publish-completeness-lib.test.mjs` — 12/12
- Green path, live against npm: all 13 packages at 1.1.31, exit 0, 1.0s
- Red path, end-to-end in a scratch repo: exit 1, `::error::` per missing
  package plus the remedy, private package correctly excluded
- Lockstep path: a manifest off `lerna.json` fails the run and is checked at
  **its own** version, not `lerna.json`'s
- `pnpm format:check` clean; `eslint` clean on every file this PR touches

## Deliberately not in scope

- **The race itself.** Serializing the npm publish across both lines needs a
  shared job-level concurrency group, and GitHub keeps only one pending run per
  group — a queued publish could be dropped. Worth its own issue.
- **The upstream misclassification.** `E409 Failed to save packument` belongs in
  a retry, not in "already published" (lerna-lite
  `is-npm-js-publish-version-conflict.js`).
- **Dist-tag verification.** The guard checks that a version exists, not that
  `latest` / `next` point at it.

Proposed in #2887, closed 2026-08-25 with an implementation that is in no branch
of this repository — hence a fresh one, with the reasoning re-derived from the
1.1.31 evidence.

[run]: https://github.com/mittwald/flow/actions/runs/34473093304
[fix]: https://github.com/mittwald/flow/actions/runs/34577464400

## Checklist

- [x] PR title is a Conventional Commit and matches the base branch (`ci:` →
      `main`)
- [x] `eslint` / `prettier` clean on the files this PR touches
- [x] No generated code affected; no user-facing strings; no visual changes

### DIFF
--- a/.github/scripts/publish-completeness-guard.mjs
+++ b/.github/scripts/publish-completeness-guard.mjs
@@ -1,0 +1,180 @@
+#!/usr/bin/env node
+// @ts-check
+/**
+ * Publish-completeness guard — fs/network shell around
+ * publish-completeness-lib.mjs.
+ *
+ * Fails (exit 1) when a publishable package does NOT resolve on the registry at
+ * the version this release shipped. Runs as the last step of `publish.yml`,
+ * after the push, the tag and the GitHub release — deliberately: `only_publish`
+ * (the remedy) publishes whatever the checked-out branch's manifests say, so
+ * failing before the push would leave the branch at N-1 while npm carries N,
+ * and the remedy would republish the wrong version.
+ *
+ * Exit codes: 0 complete · 1 hole, unverifiable or out of lockstep · 2 the
+ * guard could not run (bad `lerna.json`, no packages matched).
+ */
+import { appendFileSync, readFileSync } from "node:fs";
+import { execFileSync } from "node:child_process";
+import { dirname } from "node:path";
+import { setTimeout as sleep } from "node:timers/promises";
+import { isLernaPackageDir } from "./version-consistency-lib.mjs";
+import {
+  backoffDelays,
+  classifyStatus,
+  collectOutOfLockstep,
+  collectPublishable,
+  renderReport,
+  verdict,
+} from "./publish-completeness-lib.mjs";
+
+const registry = (
+  process.env.NPM_REGISTRY ?? "https://registry.npmjs.org"
+).replace(/\/+$/, "");
+
+const lerna = JSON.parse(readFileSync("lerna.json", "utf8"));
+const patterns = lerna.packages ?? ["packages/*"];
+
+if (typeof lerna.version !== "string") {
+  console.error("::error::lerna.json has no top-level `version`.");
+  process.exit(2);
+}
+
+// Same pathspec trick as version-consistency-guard.mjs: a git wildcard matches
+// across `/`, so `*package.json` finds every manifest at any depth; `endsWith`
+// then drops the root manifest, which is not a package.
+const entries = execFileSync("git", ["ls-files", "*package.json"], {
+  encoding: "utf8",
+  maxBuffer: 64 * 1024 * 1024,
+})
+  .split("\n")
+  .map((line) => line.trim())
+  .filter((path) => path.endsWith("/package.json"))
+  .filter((path) => isLernaPackageDir(dirname(path), patterns))
+  .map((path) => ({ path, manifest: JSON.parse(readFileSync(path, "utf8")) }));
+
+const publishable = collectPublishable(entries);
+
+if (publishable.length === 0) {
+  console.error(
+    `::error::No publishable package matched the lerna.json globs (${patterns.join(", ")}).`,
+  );
+  process.exit(2);
+}
+
+/**
+ * One registry lookup against the exact-version endpoint: 200 the version
+ * exists, 404 it does not. Anything else says nothing — see `classifyStatus`.
+ *
+ * @param {{ name: string; version: string }} pkg
+ */
+async function probe(pkg) {
+  const url = `${registry}/${encodeURIComponent(pkg.name)}/${encodeURIComponent(pkg.version)}`;
+  try {
+    const response = await fetch(url, {
+      headers: { accept: "application/json" },
+      signal: AbortSignal.timeout(10_000),
+    });
+    return {
+      presence: classifyStatus(response.status),
+      detail: `HTTP ${response.status}`,
+    };
+  } catch (error) {
+    return {
+      presence: classifyStatus(0),
+      detail: error instanceof Error ? error.message : "request failed",
+    };
+  }
+}
+
+const outOfLockstep = collectOutOfLockstep(lerna.version, publishable);
+for (const pkg of outOfLockstep) {
+  console.log(
+    `::error file=${pkg.path}::${pkg.name} is at ${pkg.version} — lerna.json says ${lerna.version}`,
+  );
+}
+
+/** @type {Map<string, import("./publish-completeness-lib.mjs").Result>} */
+const seen = new Map();
+const delays = backoffDelays();
+let outstanding = publishable;
+
+for (let attempt = 0; ; attempt++) {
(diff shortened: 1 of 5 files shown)


mfal:
Ran the guard locally against the live registry before reviewing:

```
node --test .github/scripts/publish-completeness-lib.test.mjs   12/12 pass
node .github/scripts/publish-completeness-guard.mjs             13/13 present at 1.1.33, exit 0, 0.83s
```

Package selection matches exactly — and `packages/core` is correctly excluded, which is the non-obvious one: it declares the **string** `"private": "true"`, so the truthiness read in `collectPublishable` is load-bearing, not a style choice. Good catch, and well documented.

I went looking for functional defects and found none. For the record, what I checked: the `encodeURIComponent` form (`%40mittwald%2F…`) is accepted by the registry; the retry loop runs 9 probes / 8 sleeps and leaves `seen` complete on every path; out-of-lockstep packages are probed at their own version and `renderReport` deliberately keeps the version out of the headline; and the `only_publish` remedy works end to end — version steps skipped, push/release skipped, guard still runs. That last one is where the guard is irreplaceable: in a backfill run the legitimately-already-published packages emit the *same* `already published` warning a 409 does, so the guard is the only thing that can tell them apart. The "runs last" argument holds too — aborting before the push would leave the branch at N-1 while npm carries N.

Two things I would change, both in comments.

### 1. The mechanism in `backoffDelays()` is not the CDN

> npm's registry is not immediately consistent across its CDN

The endpoint `probe()` uses is not edge-cached:

```
GET registry.npmjs.org/@mittwald%2fflow-react-components/1.1.33
  → 200, cf-cache-status: DYNAMIC
```

404s on it are not negatively cached either (two probes on a never-published version, 3s apart, no cache headers on either). What *is* cached with `cache-control: public, max-age=300` is the packument (`registry.npmjs.org/<pkg>`) — which the guard never reads.

This is an argument *for* the design: the guard is not polling against a cache it keeps warm itself. But it means the stated reason for the 305s is wrong, and that sentence is the only justification the number has. Whoever tunes it next will reason from it. Suggested replacement:

> npm's registry is not immediately consistent: the version this very job just published can still 404 on the exact-version endpoint. Not CDN staleness — that endpoint answers `cf-cache-status: DYNAMIC` and 404s on it are not negatively cached; the lag is origin-side.

### 2. Record what the budget was calibrated on

The 305s come from `flow-stylesheet` on the 1.1.31 backfill. If the origin-side lag tracks document size, that is the second-largest case, not the largest:

| | abbreviated packument | versions |
| --- | --- | --- |
| `flow-stylesheet` (the calibration) | 1.16 MB | 1508 |
| **`flow-react-components`** | **5.86 MB** | **1521** |

5x the document. If the budget ever breaks it breaks on the flagship package first — the false red this PR is explicitly built to avoid. A half-sentence naming the basis would keep the next one from being a mystery.

### Opinion, not a blocker

`unknown` reddens the run. Documented and reversible, so your call — but a 5xx or a rate limit means "we could not tell", and reddening an otherwise complete release for that carries the same cost the budget guards against. A warning plus a distinct exit path would read better.

Marginal: the guard step has no `if: always()`, so it is skipped when the push/release step fails after a successful publish — the one moment you would want to know whether npm is complete. The run is red anyway, hence marginal.

### Follow-ups (agreeing with "not in scope")

Worth noting for the write-side issue: lerna already computes the signal. `publishedPackages.push(pkg)` happens only in the success path, while the conflict branch returns `pkg` without pushing — which is why the 1.1.31 log reads `success published 10 packages` against 13 publishable ones. The discrepancy is printed, just never asserted on. And the misclassification is one line in `@lerna-lite/publish/dist/lib/is-npm-js-publish-version-conflict.js`:

```js
(ex.code === 'E409' && ex.message?.includes('Failed to save packument'))
```

Drop that branch and the 409 falls through to `process.exitCode = err.errno ?? 1` + `throw` — immediate, deterministic, at the fault. There is precedent in this repo for exactly this shape: `patches/@sigstore__sign@4.1.1.patch` already patches 409 handling in the publish path, and the retry scaffold sits in the same `catch` (the `EOTP` branch).

That would not make this guard redundant. A fixed write side only reports on packages lerna *attempted*; `collectPublishable()` enumerates from the workspace manifests independently, so it also catches a package that silently drops out of the publish set — a different failure class, and the dangerous one under fixed versioning with exact `workspace:*` pins. It would, however, let the completeness assertion for attempted publishes run **before** push/tag/release and prevent the broken release instead of detecting it, and it would let this budget shrink.

mfal (changes requested):
See https://github.com/mittwald/flow/pull/3150#issuecomment-5665522775

PR #3160, opened 2026-09-15
fix(MessageThread): restore list markers for ordered lists

TimAufdemKampe:
## Problem

`MessageThread`'s a11y fix in #2590 added `li { list-style: none }` to hide the marker on its own `<li>` message wrapper. That selector is an unscoped descendant selector, so it also strips markers from any list nested inside a message's content — e.g. an `<ol>` rendered by `Text`/`Markdown` inside a chat message or note.

`Text` restores `list-style` for `ul li` (so its own content lists survive `globals.scss`'s unrelated `ul { list-style-type: none }` reset), but has no equivalent rule for `ol`, and outside of `MessageThread` it never needed one — native `ol` numbering just works. So the combination is chat-specific: bullets in unordered lists survive because `.text ul li` (specificity 0,1,2) beats `.messageThread li` (0,1,1), but ordered list numbers have no such rule to win the fight and get silently killed only inside a message thread.

(Earlier revision of this PR tried fixing it in `Text` with `list-style: initial` for `ol` too — that turned out to just render `disc` bullets instead of numbers, since `initial` resets to the CSS spec's initial value, not the element's native default. Reverted in favor of the actual root cause below.)

## Fix

Scope `MessageThread`'s reset to `> li` — it only ever needs to hide the marker on its own direct `<li>` message wrapper, not on lists nested arbitrarily deep inside message content. `Text` needs no change: `ol` already renders correctly everywhere once `MessageThread` stops reaching into it.

## Test plan

- [x] Ordered list rendered inside `Text` standalone still shows numbers
- [x] Ordered list rendered inside `Text` nested in `MessageThread` now shows numbers (previously invisible, then briefly bullets)
- [x] Unordered list bullets unaffected
- [x] Visual regression baselines updated and reviewed


ins0:
@TimAufdemKampe a Visual Test that covers this behavior would be nice!

ins0 (approved):

PR #3166, opened 2026-09-15
feat(CoachMark): add a self-opening hint that points at a control

### LINKED ISSUE #3165: ContextualHelp: no non-modal variant that is still announced — modal locks page scrolling, isNonModal strips all semantics
`ContextualHelp` today offers two states, and for an informational popover that opens without user action, both are wrong in a different way. There is no middle ground: **non-modal but still announced.**

## The use case

An anchored hint that opens on mount and points at a newly introduced control — a feature spotlight. It is informational, has a single "Got it" button, and is not something the user asked for. Rendered as:

```tsx
<ContextualHelp controller={controller} placement="bottom" triggerRef={triggerRef} width={400}>
  <Section>
    <Heading>{title}</Heading>
    <Text>{description}</Text>
    <Action onAction={acknowledge}><Button size="s">Got it</Button></Action>
  </Section>
</ContextualHelp>
```

## What the two states do

Measured in a real app (Flow `1.1.38`, `react-aria-components` `1.20.0`), on a page ~3200px tall in a 1324px viewport:

| | default | `isNonModal` |
| --- | --- | --- |
| Popover renders | yes | yes, correctly positioned |
| Page scrolling | **blocked** — `<html style="…; overflow: hidden">`, mouse wheel and PageDown both dead | works normally |
| Role | `role="dialog"` | **none** — no role and no aria attribute anywhere on the chain |
| Focus | moves into the popover | stays on `body` |
| Button reachable by Tab | yes | yes |
| Background `aria-hidden` | **no** | no |

The scroll lock is react-aria's overlay behaviour and correct for a modal. The problem is that an uninvited informational hint should not be modal at all — until it is dismissed, the user cannot read the page it is pointing at, with either mouse or keyboard. Note the default does not even hide the background from assistive tech, so the modality buys the lock without the containment that would justify it.

`isNonModal` fixes the lock, but the popover then carries no semantics at all: nothing announces it, and focus never enters it. For a hint whose entire purpose is "notice this", that is worse than the lock — the users who most need an explicit pointer are the ones who would never hear about it.

## The ask

A variant or prop for **non-modal but announced**: no scroll lock and no focus trapping, but keep the dialog semantics (role + labelling from the `Heading`), and optionally either move focus deliberately or announce via `aria-live`. Naming is yours — something like `modality="modal" | "non-modal"` on `ContextualHelp`, or a dedicated hint variant.

Two smaller things that came up on the way, take or leave:

- `isNonModal` currently reaches `Aria.Popover` only because `PopoverProps extends Omit<Aria.PopoverProps, …>` spreads everything through. It is not documented at the Flow level and, as the table shows, silently strips the semantics — either way it goes, documenting what the two modes do would help.
- If a non-modal variant lands, it would be worth saying in the docs which one a `ContextualHelp` that opens itself should use.

Context: we hit this building a reusable feature-spotlight mechanism in mStudio; the popover anchors via `triggerRef` without a `DialogTrigger` ancestor, which works fine in both states.

mfal:
A hint that appears on its own must not lock the page, and must not vanish on the first scroll. `ContextualHelp` could do neither.

The semantics differ too. A `ContextualHelp` is a dialog you open: it is announced, and it takes focus. A `CoachMark` is plain content on the page — no role, no announcement, no focus.

That is why it stays where it is written, right behind its anchor, instead of being portalled to the end of the `<body>` the way a dialog is. And why it is its own component rather than another prop on `ContextualHelp`.

## `CoachMark`

Points at a control that is new, moved or easy to miss. It opens on its own, so it stays out of the way:

| | `ContextualHelp` | `CoachMark` |
| --- | --- | --- |
| Who opens it | the user, via a trigger | the app, unasked |
| Page scrolling | blocked | works, and it rides along with its anchor |
| Background | `inert` | untouched |
| Focus | moves into it | stays where it was |
| Semantics | `dialog`, announced | plain content |
| Dismissing | close on outside press | its own button, localized, plus Escape |

The anchor is an `anchorRef`, not a `triggerRef` — nothing triggers it.

Ships as **beta** (`@flowStatus beta, new`), with a docs page under Overlays and a note that the API can still change.

`ContextualHelp` is untouched: `modality` is gone from its props again, and it is exactly what it was before this PR.

## Anchoring by `id`, so an extension can place one

An mStudio extension renders in a different context than the host. A ref never crosses that boundary; a string does. So the anchor can also be given as the anchor element's `id`, which the host looks up when the coach mark opens:

```tsx
<Button id="deploy">Deploy</Button>
<CoachMark anchor="deploy">…</CoachMark>
```

`anchorRef` stays for the local case and is cut from the remote surface with `@flr-ignore-props` — it could not work there. `id` needed no infrastructure change: it is already a remote property.

That makes the component `@flr-generate all`, with the generated artifacts committed and a demo page at `/remote/coach-mark` in `apps/remote-dom-demo`. The visual scenario runs in **both** environments against the one shared baseline, so an anchor that fails to resolve on the host shows up as a diff rather than as nothing.

Four props are `Omit`ted from the inherited `PopoverProps`, because a coach mark cannot honour them: `getTargetRect`, `shouldCloseOnInteractOutside`, `isKeyboardDismissDisabled`, `UNSTABLE_portalContainer`. The generator rejects the first two anyway — a function prop travels as a proxy and returns a Promise to the host — but they are meaningless locally too, so they come out of the type rather than just out of the remote surface.

## The non-modal path does not use `usePopover`

react-aria's `usePopover` couples three things: it locks page scrolling, it hides the rest of the page from assistive tech, and it closes the popover on the first scroll — measured, ~150px was enough to lose a hint before it had been read. There is no opt-out, and no `shouldCloseOnScroll` exists in `react-aria@3.52.1`.

`NonModalPopoverContent` therefore builds on `useOverlayPosition`, public react-aria API, and keeps none of it. Surviving a scroll is a property of the construction, not a flag. What `Aria.Popover` and `Aria.OverlayArrow` contribute on the modal path is reproduced: `data-placement` (the SCSS keys the tip direction off it), the entry animation, and the tip's absolute placement from `arrowProps`.

`modality` stays on `Popover` as the general capability this is built on.

## It renders where it stands

A portal to the end of `<body>` is right for a modal popover: focus moves into it and everything else goes `inert`, so its place in the document never comes up. A coach mark is announced by nothing and has to be found by reading — at the end of the body it would sit behind the whole page. So it is not portalled, and reading order is anchor, hint, rest of the page. On top of that the anchor refers to it with `aria-details`.

`useOverlayPosition` resolves the containing block through `offsetParent`, so the anchoring is unaffected. The cost: it inherits its surroundings — an ancestor that clips or opens a stacking context can cut it off. Documented.

## A popover that opens itself has to keep looking

`useOverlayPosition` recalculates on window resize, on scroll of a scroll container, and when the anchor or the popover changes *size*. It never notices the anchor merely *moving* — which is fine for a popover opened by a user, long after the page settled.

A coach mark places itself into a page that is still filling in. In the remote demo the host's header arrives once the connection is up, pushes the anchor down, and leaves the coach mark 74px above it, covering the button it points at. Remotely that is the normal case, not an edge case: the host materializes the page in pieces as they arrive.

So the non-modal popover observes the anchor's ancestors and repositions. The anchor is read inside the effect, not during render: a coach mark that is open from its first render renders before the anchor's ref is attached, and a ref filling in re-renders nothing, so reading it during render saw `null` and left nothing observing. In block layout an element moves because something around it changed size, and a resized ancestor is the signal every such reflow has in common. Verified over a real iframe connection in `apps/remote-dom-demo`: on first paint, without any interaction, the coach mark centres on its anchor and sits 16px below it. The regression test fails by exactly the injected 200px without the fix.

## Six defects the docs page and the remote demo turned up

Verifying the docs and the remote demo surfaced three bugs that only show under
conditions the earlier checks never hit. Each has a regression test that fails
without its fix.

**The tip sat on the wrong edge when the popover was placed above its anchor.**
`arrowProps.style` carries the offset along the popover's edge and leaves the
other axis `undefined` — which still overwrites when spread. Spread after the
`[placement]: "100%"` that puts the tip on an edge, it erased `top: 100%`, and
the tip fell back to the popover's top edge, pointing away from the anchor.
`placement: "bottom"` was never affected, because there the key is `bottom` and
nothing overrode it — which is exactly why every earlier check passed.

**An unplaced popover was painted in the top left corner.** Until it has measured
the anchor, `useOverlayPosition` parks the overlay at `position: fixed; top: 0;
left: 0`. A coach mark opens itself, so a server renders it open and that corner
is what ships: the hint sits there until hydration measures, then jumps to its
anchor. It now stays invisible until it has a placement, and in the DOM
throughout — its reading order and `aria-details` depend on that.

**It claimed `z-index: 100000` and scrolled over the page's own header.** That
number is what an overlay portalled to the end of `<body>` needs to clear the
page. Rendered in place it put the hint in front of the whole application. It now
claims no stacking level: it still covers the ordinary content around it, and app
chrome that sets a `z-index` keeps its place above.

**A coach mark below the fold rendered as a two-pixel sliver.** `max-height` is
the room left in the viewport at the moment of measuring, and an anchor below the
fold has none — page scrolling never triggers a recompute, so the clamp stuck
however far the user scrolled. The hook writes that value straight onto the
element rather than returning it among its props, so the override lives in CSS.

**`onOpenChange` suppressed the close it reported, and never saw the dismiss button at all.** Filed as #3174 and fixed on its own in #3183, which also gave `Popover` a real controlled mode and renamed `defaultOpen` to `isDefaultOpen`. This branch is rebased onto that fix: `onOpenChange` is registered on the controller now, so it reports every close path, the dismiss button included. `CoachMark` ships only `isDefaultOpen` — it has never been released, so `defaultOpen` joins its `Omit` list instead of arriving deprecated on day one, and the codemod that carries the rename for `Popover`, `ContextualHelp` and `ContextMenu` correctly leaves it alone.

**`anchorRef={condition ? ref : undefined}` broke the anchor entirely.**
`getPropsMerger` treats every prop ending in `Ref` as a ref and merges what it
collects; it filtered out `null` but kept `undefined`, so a prop that was passed
but unset still reached `mergeRefs` and came back as a callback ref. Harmless
for a ref the component attaches, fatal for one it only reads: nothing calls the
callback, `.current` stays empty, and the anchor is never found. A real ref was
never affected.

## Deliberately not announced

#3165 asked for non-modal **plus** dialog semantics, so a self-opening hint gets announced. That is not what shipped: the consumer that needs this wants the content presented as ordinary content, without a role and without a focus grab. The docs state the trade-off. A polite live region is the mechanism if that changes; it is orthogonal and additive.

## Regression caught on the way

react-aria hands `Aria.Popover` an `isNonModal` through **context** — `ComboBox` sets it so the user can keep typing. `PopoverContent` therefore only decides the modality when someone actually asked for one; an explicit `false` would have turned every combo box option list into a modal that locks the page. `ComboBox.browser.test.tsx` guards it, verified to fail on the buggy version.

## Verification

- `test:compile`, `test:unit` and `test:browser` (webkit, what CI runs) green across the workspace; the CoachMark visual scenario green in both environments.
- Checked live in Storybook: opens on load, dismiss button localized, no `dialog` role, `html` unlocked, focus untouched, `aria-details` resolves, and the offset to the anchor stays exactly constant across 0 → 500 → 0 px of scrolling.
- The `-webkit-linux` baselines are committed. They were taken while the scenario was Local-only; the rendering has not changed since, only the way the anchor is found, so they should still hold — CI on this head is the confirmation.

## Notes for reviewers

- `@flr-generate` and `flr-universal.ts` are **either-or**, not both. `remote-react-components/src/index.ts` re-exports `./auto-generated` and `./components` with two `export *`, and ESM drops a name that both provide — silently. CoachMark was in both for a while and arrived on the host as `undefined`, with React reporting only "Element type is invalid". So CoachMark is in `public.ts` only; adding it to `flr-universal.ts` would break it again. (`Action`, `LightBox` and `List` are the other way round: in `flr-universal`, absent from `auto-generated`.) Now an entry in the AGENTS.md footgun list.
- The docs page carries no beta notice of its own: the docs site renders one from the component status registry, and the hand-written `<Alert>` said the same thing directly below it.
- `src/components/Option/Option.browser.test.tsx` fails on **firefox** in a full local run. It fails identically on the untouched base commit `d810b78`, and CI runs webkit only.

fixes #3165


apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+---
+
+# Best Practices
+
+- Setze ein CoachMark nur für einen Hinweis ein, um den niemand gebeten hat.
+  Alles, was Nutzende selbst aufrufen, gehört in ein
+  [ContextualHelp](/components/overlays/contextual-help).
+- Zeige höchstens ein CoachMark pro Seite. Mehrere konkurrieren um dieselbe
+  Aufmerksamkeit und heben sich gegenseitig auf.
+- Beschränke den Inhalt auf einen Gedanken. Sag, was neu ist und was es bringt.
```
Jan-Eimertenbrink: Zu umgangssprachlich geschrieben formuliere noch mal neu Sachlich und empfehlend

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+  oder leicht zu übersehen ist.
+---
+
+<LiveCodeEditor />
+
+---
+
+# Best Practices
+
+- Setze ein CoachMark nur für einen Hinweis ein, um den niemand gebeten hat.
```
Jan-Eimertenbrink: Der erste Satz ann komplett raus, da direkt darunter im Vs bereich noch mal der unterschied deutlich wird

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
+---
+component: CoachMark
+description:
+  Ein CoachMark weist ungefragt auf ein Bedienelement hin, das neu, verschoben
+  oder leicht zu übersehen ist.
```
Jan-Eimertenbrink: Eine CoachMark macht auf neue Features oder verschobene Bedienelemente Aufmerksam.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+# Best Practices
+
+- Setze ein CoachMark nur für einen Hinweis ein, um den niemand gebeten hat.
+  Alles, was Nutzende selbst aufrufen, gehört in ein
+  [ContextualHelp](/components/overlays/contextual-help).
+- Zeige höchstens ein CoachMark pro Seite. Mehrere konkurrieren um dieselbe
+  Aufmerksamkeit und heben sich gegenseitig auf.
+- Beschränke den Inhalt auf einen Gedanken. Sag, was neu ist und was es bringt.
+- Blende ein bestätigtes CoachMark dauerhaft aus. Ein Hinweis, der bei jedem
```
Jan-Eimertenbrink, line 18: Auch hier ist es zu Umgangssprachlich formuliert. Eher so: Blende ein bestätigtes CoachMark dauerhaft aus. Ein Hinweis, der bei jedem Aufruf erneut erscheint, kann als störend wahrgenommen werden.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+  Aufruf wiederkommt, wird zur Störung.
+- Verlasse dich nicht darauf, dass der Hinweis gelesen wird. Was unerlässlich
```
Jan-Eimertenbrink: Platziere unerlässliche Informationen direkt in der Oberfläche. Ein CoachMark sollte nicht die einzige Quelle für wichtige Informationen sein.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+  ist, gehört in die Oberfläche selbst.
+- Schreibe das CoachMark direkt hinter sein Ankerelement. Angekündigt wird es
```
Jan-Eimertenbrink: Platziere das CoachMark direkt hinter seinem Ankerelement. So wird es in der Vorlesereihenfolge an der passenden Stelle angekündigt.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+  nicht, gefunden wird es über die Vorlesereihenfolge.
+- Nutze in einer mStudio-Extension die `anchor`-Property. Eine Referenz erreicht
```
Jan-Eimertenbrink: Das könnte eigentlich ein teil des darüberliegeneden Stichpunkts sein. Normalerweise gehen wir in Flow auch nicht zu tief auf Besonderheiten der mStudio Extensions ein.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+  den Host nicht, eine id schon.
+
+## CoachMark vs. ContextualHelp
+
+Beide zeigen einen kurzen Hinweis neben einem Bedienelement. Der Unterschied
+liegt darin, wer ihn öffnet – und was das für die restliche Seite bedeutet. Ein
+ContextualHelp ist ein Dialog: Es sperrt das Scrollen und zieht den Fokus zu
+sich. Ein CoachMark lässt die Seite unberührt.
+
```
Jan-Eimertenbrink, line 32: Verwende eine CoachMark um den Nutzer proaktiv auf ein neues oder verschobenes Feature aufmerksam zu machen. Für weiterführende Informationen ist Contextual Help die bessere Wahl, da der Nutzer selbst entscheiden kann, ob er diese Information braucht oder nicht.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+<DoAndDont>
+  <Plain heading="Verwende ein CoachMark, um z. B. ...">
+    - auf eine neue oder verschobene Funktion hinzuweisen.
+    - eine Funktion zu zeigen, die leicht zu übersehen ist.
+    - einen Hinweis zu geben, ohne die Arbeit zu unterbrechen.
+
+  </Plain>
+
+  <Plain heading="Verwende ein ContextualHelp, um z. B. ...">
+    - einen Fachbegriff zu erklären, wenn jemand danach fragt.
```
Jan-Eimertenbrink: > einen Fachbegriff zu erklären 

Reicht der nebensatz kann weggelassen werden

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+    - eine Einstellung mit ihren Folgen zu beschreiben.
+
+  </Plain>
+</DoAndDont>
+
+---
+
+# Anker
+
+Ein CoachMark zeigt auf ein Element, nicht auf die Seite. Einen Trigger gibt es
```
Jan-Eimertenbrink, line 53: Ein CoachMark zeigt auf ein Element, nicht auf die Seite. Einen Trigger gibt es nicht, da das CoachMark nicht vom Nutzer geöffnet wird. Stattdessen wird sein Ankerelement auf einem von zwei Wegen angegeben:

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+nicht, denn niemand öffnet es – stattdessen bekommt es den Anker auf einem von
+zwei Wegen:
+
+- `anchorRef` ist die Referenz auf das Element. Das ist der Normalfall in einer
+  React-Anwendung.
+- `anchor` ist die `id` des Elements, nachgeschlagen beim Öffnen. Sie hilft
+  dort, wo sich keine Referenz teilen lässt: Eine mStudio-Extension rendert in
+  einem anderen Kontext als der Host, eine Referenz kommt dort nie an, eine id
+  schon.
+
```
Jan-Eimertenbrink: der nachfolgende Abschnitt bis zum editor ist etwas lang. Vielleicht eher so um ihn deutlich in der Länge zu kürzen: 

Das CoachMark wird direkt hinter seinem Ankerelement im DOM gerendert und folgt diesem in der Vorlesereihenfolge. Über aria-details verweist der Anker zusätzlich auf das CoachMark. Es wird nicht automatisch angekündigt, sondern beim Lesen erreicht.

Das CoachMark ist absolut positioniert und benötigt keinen eigenen Platz. overflow: hidden oder ein eigener Stacking Context eines Vorfahren kann es abschneiden.

Beim Scrollen folgt das CoachMark seinem Anker und bleibt geöffnet.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+abschneiden.
+
+Beim Scrollen wandert das CoachMark mit seinem Anker mit und bleibt geöffnet.
+
+<LiveCodeEditor example="anchorById" editorCollapsed />
+
+---
+
+# Schließen
+
```
Jan-Eimertenbrink, line 78: Das CoachMark enthält standardmäßig einen Button zum Schließen. Über dismissLabel lässt sich die Beschriftung anpassen, mit hideDismissButton kann der Button ausgeblendet werden. Auch Escape schließt das CoachMark.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Übernommen als `description`: „Ein CoachMark macht auf neue Features oder verschobene Bedienelemente aufmerksam." — im Neutrum, weil die Seite und deine übrigen Kommentare durchgehend „das CoachMark" schreiben.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Erster Satz raus. Der Bullet ist jetzt nur noch der Verweis: „Nutze für Informationen, die der User selbst abruft, ein ContextualHelp."

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Neu und sachlich: „Beschränke den Inhalt auf einen Gedanken. Benenne, was neu ist und welchen Vorteil es bringt." („Sag, was …" war das Umgangssprachliche.)

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Wörtlich übernommen. Den Mechanismus (`onOpenChange` / `isDefaultOpen`) hatte ich hier stehen — der ist jetzt in `# Schließen`, wo er hingehört.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Wörtlich übernommen.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Übernommen, mit einer Abweichung: „erreicht" statt „angekündigt". Ein CoachMark wird bewusst nicht angekündigt — deine eigene Formulierung im Anker-Thread sagt „nicht automatisch angekündigt, sondern beim Lesen erreicht". So bleiben beide Stellen widerspruchsfrei.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Zusammengeführt. Der Anker-Bullet endet jetzt mit „Steht keine Referenz zur Verfügung, gib den Anker als `id` an." — ohne mStudio. Die Begründung steht weiterhin in `# Anker`.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Übernommen als Einleitung des Vergleichs; die mechanische Modal/Non-Modal-Erklärung ist raus.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Nebensatz gestrichen: „einen Fachbegriff zu erklären."

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Wörtlich übernommen.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Wörtlich übernommen, die drei Absätze ersetzen die bisherigen vier. `aria-details` und `overflow: hidden` stehen als Inline-Code, wie die Guideline es für Attribute und CSS vorgibt.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Übernommen. Einzige Ergänzung: „beschriftet mit einem übersetzten „Verstanden"" bleibt im ersten Satz — dass das Label schon lokalisiert ist, geht aus dem deutschen Beispiel allein nicht hervor. Sag Bescheid, wenn es trotzdem raus soll.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+  aufmerksam.
+---
+
+<LiveCodeEditor />
+
+---
+
+# Best Practices
+
+- Nutze für Informationen, die der User selbst abruft, ein
```
Jan-Eimertenbrink: Der erste Stichpunkt kann hier raus, da es im Vs Bereich direkt darunter genauer und besser erklärt wird.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
```diff
@@ -0,0 +1,96 @@
...
+  sollte nicht die einzige Quelle für wichtige Informationen sein.
+- Platziere das CoachMark direkt hinter seinem Ankerelement. So wird es in der
+  Vorlesereihenfolge an der passenden Stelle erreicht. Steht keine Referenz zur
+  Verfügung, gib den Anker als `id` an.
+
+## CoachMark vs. ContextualHelp
+
+Verwende ein CoachMark, um den User proaktiv auf ein neues oder verschobenes
+Feature aufmerksam zu machen. Für weiterführende Informationen ist ein
+ContextualHelp die bessere Wahl, da der User selbst entscheiden kann, ob er
```
Jan-Eimertenbrink: Verlinke das ContextualHelp

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Bullet komplett raus. Der ContextualHelp-Link ist in den Vergleich darunter gewandert, damit die Component von der Seite aus erreichbar bleibt.

apps/docs/src/content/components/overlays/coach-mark/index.mdx
mfal: Verlinkt: „Für weiterführende Informationen ist ein [ContextualHelp](/components/overlays/contextual-help) die bessere Wahl …"

ins0:
Review notes — the four findings I would fix before merge. CI on `e59c6d9` is green and the generated code is reproducible (`pnpm nx build stylesheet` in a clean worktree leaves `git status` empty), so these are behavioural. All four were verified with throwaway probe tests in WebKit, not read off the diff.

### 1. A late-arriving anchor is never found — `anchor` resolves once, at mount

`CoachMark.tsx:57`: `useEffect(…, [anchor])`. The deps are the string alone, so `document.getElementById` runs once after the first commit and never again.

Verified: anchor mounted 100ms after the coach mark → 800ms later the popover still sits at `position: fixed; top: 0; left: 0; visibility: hidden` and the anchor has no `aria-details`, although `getElementById` now finds it. No retry, no dev-time signal — the hint is silently invisible.

The documented contract says otherwise in four places:

- `CoachMark.tsx:39` — "looked up when it opens"
- `CoachMark.tsx:48` — "The id is looked up after the commit, so the anchor is in the DOM by then"
- `apps/docs/src/content/components/overlays/coach-mark/index.mdx:59` — "nachgeschlagen beim Öffnen"
- `apps/remote-dom-demo/src/app/remote/coach-mark/page.tsx:13` — "The host looks the element up once the coach mark opens"

And it is the case `anchor` exists for: the description argues the host "materializes the page in pieces as they arrive". Repositioning a moving anchor is solved; resolving one that appears later is not. Minimum fix: `isOpen` in the deps. Better: keep looking until it resolves, and make the miss observable.

### 2. `"A coach mark without any anchor renders nothing"` cannot fail, and claims the opposite of what happens

`CoachMark.browser.test.tsx:369`. `expect.element(...).not.toBeInTheDocument()` straight after `render()` passes on the poll's first attempt, before React commits. Control: the same assertion against `<Text data-testid="hint">I am definitely here.</Text>` is green too. The test would pass for any component.

The real behaviour is that it renders — full popover with content, `visibility: hidden`, parked at the origin. `useAnchorRef` always returns a ref object, so the `if (!anchor) return null` guard at `PopoverContent.tsx:67` is dead code for `CoachMark`. Every other negative assertion in the file awaits presence first, which is why those hold.

### 3. `maxHeight` is on the prop type and on the remote surface, and does nothing

`Popover.module.scss:16` `max-height: none !important` defeats it. Verified: `<CoachMark maxHeight={60}>` → inline `max-height: 60px`, computed `none`, rendered 162px. It is absent from the `Omit` list, ships as a remote property (`RemoteCoachMarkElement.ts:33`), and appears in the docs page's `<PropertiesTables />`. Either omit it like the other four and regenerate, or scope the override so a value the consumer set wins.

### 4. DOM props are dropped on the non-modal path

`NonModalPopoverContent.tsx:97` destructures a fixed set and funnels everything else into `positionProps`, which only reaches `useOverlayPosition`. Verified lost: `aria-label`, `data-*`, `style`. `Aria.Popover` forwards them on the modal path. `aria-label` / `aria-labelledby` / `aria-describedby` / `aria-details`, `dir`, `hidden`, `inert`, `lang`, `slot` and `translate` are generated remote properties, and `style` is the one generated remote attribute — so an extension can set them and nothing happens.

3 and 4 are the other half of the cleanup the four new `Omit`s started.

mfal:
All four reproduced before fixing, all four fixed in `08879c622`. Thanks — two of them were invisible from my side precisely because the tests said otherwise.

**1. Late anchor.** Confirmed exactly as described: anchor mounted 100ms late → `style=position: fixed; top: 0px; left: 0px; visibility: hidden`, `anchorInDom=true`, `ariaDetails=null`. `useAnchorRef` now keeps looking with a `MutationObserver` on `document.body` and disconnects the moment it resolves. `isOpen` in the deps would only have covered an anchor that arrives before the open; the observer covers one that arrives at any point, which is what the four doc comments promised.

**2. The vacuous test.** Confirmed, including your control — the assertion passes for anything. Replaced by what actually happens: without an anchor the coach mark renders, parked and `visibility: hidden`, and the test waits for presence first. Added a second test for the late anchor, verified to fail without the observer (`expected undefined to be truthy`).

The dead guard at `PopoverContent.tsx:67` stays as it is: `useAnchorRef` always returns a ref object, so `CoachMark` never reaches it, but a `Popover` used with `modality="non-modal"` and no trigger still can.

**3. `maxHeight`.** Confirmed: `inline=60px computed=none height=82`. It joins the `Omit` list. Scoping the override instead is not reachable — react-aria writes one `max-height` that is already `Math.min(userSetMaxHeight, viewportDerived)`, so CSS cannot tell a consumer's value from the clamp this component has to escape.

**4. DOM props.** Confirmed: `aria-label=null data-testid=null lang=null`. The positioning props are named one by one now and the remainder is spread onto the element. Regression test verified to fail without the spread.

Gate: lint, `test:compile` across the workspace, 342 browser tests (webkit), unit, visual 366/366, `pnpm build` with no generated drift. `RemoteCoachMarkElement` lost `maxHeight`, which is the only generated change.

ins0 (approved):

Issue #3189, opened 2026-09-18
Action: closeOverlay without a name should close the nearest overlay


mfal:
`<Action closeOverlay>` should close whatever overlay encloses it. Today it does not compile — `closeOverlay` takes a name, a controller or options, and a bare attribute is `true`:

```
src/components/Action/types.ts:25:3 - The expected type comes from property 'closeOverlay'
  closeOverlay?: FlowComponentName | OverlayController | CloseOverlayOptions;
```

So every composed dismiss action has to name the overlay it sits in:

```tsx
<CoachMark anchor="deploy">
  <Text>…</Text>
  <Action closeOverlay="CoachMark">
    <Button>Verstanden</Button>
  </Action>
</CoachMark>
```

The name is redundant — the action is *inside* that overlay — and it leaks the composition: a consumer has to know which Flow component the surrounding overlay is, which is not always the one they wrote (`ModalTrigger` registers as `Modal`, a `CoachMark` was reachable only as `Popover` until it started registering its own name).

## Why it is not just a missing union member

`closeOverlay` resolves against a flat map keyed by component name:

```ts
export type OverlayContext = Partial<
  Record<FlowComponentName, OverlayController | undefined>
>;
```

`OverlayContextProvider` builds it as `{ ...parentContext, [type]: controller }`. So "innermost wins" already holds **per name** — a Modal inside a Modal overwrites the `Modal` key — but nothing records which provider is innermost **across** names. A `Popover` inside a `Modal` leaves both keys set, in no particular order, and `ActionModel.getOverlayController` has nothing to pick from:

```ts
if (typeof from === "string") return this.overlayContext[from];
if ("overlay" in from)        return this.getOverlayController(from.overlay);
return this.getOverlayController("Modal");   // today's fallback
```

Note the fallback: anything truthy that is neither a name nor `{ overlay }` already resolves to `"Modal"`. A boolean would silently close the wrong thing rather than the nearest one — and `"overlay" in true` throws before it gets there.

So the context has to carry the nearest controller, e.g. a dedicated slot every provider writes alongside its own key. Four components provide one today: `OverlayTrigger` (with the trigger's `overlayType`), `Overlay` (as `Modal`), `Popover`, and `CoachMark`.

## Decisions this needs

[...]

PR #3191, opened 2026-09-18
feat(Action): close the nearest overlay with a bare closeOverlay

mfal:
`<Action closeOverlay>` now closes whatever overlay encloses it. A composed dismiss action no longer has to name the surrounding component — a name that leaked the composition, because it is not always the component the consumer wrote (`ModalTrigger` registers as `Modal`).

```tsx
<CoachMark anchor="deploy">
  <Text>…</Text>
  <Action closeOverlay>
    <Button>Verstanden</Button>
  </Action>
</CoachMark>
```

## Why the name was required

Collateral damage of #506 (*"fix overlays are toggled by unrelated triggers"*). Before it, the overlay context was a single controller — implicitly the nearest one — and the prop was `closeOverlay?: boolean | OverlayController`:

```ts
export const overlayContext = createContext<OverlayController | undefined>(undefined);
```

That provider handed its controller to every descendant, so a `ContextMenuTrigger` inside a `Modal` opened along with the modal. #506 keyed the context by component name to fix it, and dropped the nesting information in the same step — which left `true` with nothing to mean, so all three props became `FlowComponentName | OverlayController`.

## What changed

The context carries both pieces now:

- **`byType`** — the map from #506, unchanged. Still what `useOverlayController` resolves a name against, so the fix stays intact.
- **`nearest`** — the innermost overlay whose *content* the subtree is, whatever its component name.

The decisions the issue asked for:

- **The provider around the trigger.** `OverlayTrigger` wraps the trigger as well as the overlay, and the trigger is not inside the overlay it opens. It passes `isOverlayContent={false}` and contributes to `byType` only. Nothing is lost: the content components (`Overlay`, `Popover`, `ContextMenu`) re-provide the same controller and set `nearest` there.
- **An explicit name keeps winning.** `closeOverlay="Modal"` inside a `Popover` still closes the outer modal.
- **No `openOverlay={true}` counterpart.** The nearest overlay is the one the action renders inside, so it is open by definition; `toggleOverlay={true}` would be a confusing alias for closing. The asymmetry is deliberate and documented on the prop.
- `CloseOverlayOptions.overlay` became optional and means the same as `true`.

`getOverlayController` is also split into the *"does this action carry an overlay prop at all"* guard that `Action` depends on for the surrounding `Modal`'s controller, and the reference resolution itself. The old nested closure switched on its own parameter but resolved the outer `from`.

## Call-sites

`ContextMenu` and `ContextMenuSection` were the only places naming their own overlay. Their `Action` sits directly inside the menu's overlay context, so the name carried no extra information. `getCloseOverlayType` existed only to map the selection mode onto that name or `undefined`; it is really a yes/no question and becomes `closesOnSelect`, returning a boolean.

`Modal.browser.test.tsx` keeps its `closeOverlay="Modal"` on purpose — that case asserts the named form still bypasses the close confirmation.

## Cross-version

The in-process harness went red on the `next` line for two scenarios unrelated to this change, both pre-existing divergences rather than regressions (the iframe harness passed):

- **CartesianChart Bars** — `Bar` landed in `1.2.0-next.6`; the tested `1.1.49` and `1.2.0-next.0` resolve it to `undefined` and the host dies with *Element type is invalid*. Now gated with `below: barSince`.
- **ImageCropper custom error view** — already gated `below: "1.1.41"`, but the `next` line branched before #3024 forward-merged. The fix lands in `1.2.0-next.50`, so all 50 versions from `1.2.0-next.0` to `next.49` sort *above* `1.1.41` without carrying it — which `below` cannot express. Gated with a new `excludeRange: ">=1.2.0-next.0 <1.2.0-next.50"`, which states what was published rather than matching one target selector's current output.

  The rules moved to `skipsVersion.ts` so they are reachable from a unit test — `environments.ts` imports the old published packages and dies at import time outside the prepared harness. Twelve cases pin the classification across the line boundary.

`crossVersion` is a no-op in the normal visual suite, so the Local/Remote environments and their baselines are untouched.

## Compatibility

Additive only — every existing form of the prop keeps working. `Action` is not `@flr-generate`, so there is no generated code to regenerate.

## Tests

`src/components/Action/closeNearestOverlay.browser.test.tsx` covers the nearest modal, the nearest popover, inner-wins-over-outer, an explicit name still winning, and a trigger's subtree not counting as inside the overlay it opens. `test:compile` and the Action/Modal/ContextMenu/List/controller browser suites (102 tests) pass on webkit; the visual suite ran green via `run-visual-tests`.

fixes #3189


ins0:
Reviewed this branch. The feature holds: I mutation-tested the new suite (pointing `resolveOverlayReference`'s `nearest` branch at `undefined` fails 4 of 5; removing `isOverlayContent={false}` fails exactly the trigger test), `pnpm build` leaves no generated-code drift, and `test:compile` passes for both touched packages. One finding, in the cross-version gating rather than the feature.

## `ImageCropper` — the `exclude` comment understates the hole

The gate is correct: every version the selector can currently reach is either excluded or carries the fix. What's wrong is the comment, which presents `1.2.0-next.0` as the whole hole.

```bash
git merge-base --is-ancestor 28fa2842b 1.2.0-next.49 && echo has || echo missing
```

`1.2.0-next.0` … `1.2.0-next.49` all lack the fix; it lands in `1.2.0-next.50`. The single entry only suffices because `firstOfLine` is the one selector that reaches into that range — at this branch's `1.2.0-next.59`, `selectCrossVersionTargetVersions` returns `previous=1.2.0-next.58`, `firstOfLine=1.2.0-next.0`, `latestOfPreviousLine=1.1.51`.

Whether it recurs: every other gate in the suite thresholds on `0.2.0-alpha.*` (pre-split, no hole possible), `ratingSegmentSince = "1.1.0"` predates the branch point, and `barSince` is a next-line threshold with nothing on the 1.1 line to miss — and it's exact (`1.2.0-next.5` has no `Bar`, `next.6` does). So this is the only instance today, but it's the first case of a structural pattern: any `main`-line fix a scenario depends on will forward-merge late.

### Minimum fix — correct the comment

```ts
// errorView is available from 1.1.41 on the 1.1 line. The next line branched
// before #3024 forward-merged, so 1.2.0-next.0 … 1.2.0-next.49 all sort above
// 1.1.41 without carrying it (it lands in 1.2.0-next.50). Only next.0 is listed
// because `firstOfLine` is the only target selector that reaches that range —
// see dev/cross-version/selectTargetVersions.ts.
test
  .skipIf(crossVersion({ below: "1.1.41", exclude: ["1.2.0-next.0"] }))
```

### Better fix — express the range

Four lines, and the gate stops being right-by-argument. `src/tests/lib/environments.tsx` and `e2e/cross-version-inprocess/environments.ts` declare `CrossVersionSkip` separately, so both need the field:

```ts
export interface CrossVersionSkip {
  /** Skip when the tested version is older than this (semver). */
  below?: string;
  /** Skip these exact versions (for non-monotonic breakage). */
  exclude?: string[];
  /**
   * Skip every version in this semver range. For a fix that landed on `main`
   * and forward-merged to `next` later: the versions published on the next line
   * in between sort above the `below` threshold without carrying it.
   */
  excludeRange?: string;
}
```

And the predicate in the e2e copy:

```ts
export const crossVersion = ({
  below,
  exclude,
  excludeRange,
}: CrossVersionSkip): boolean => {
  const version = __FLOW_CROSS_VERSION__;
  if (semver.valid(version) === null) {
    return false;
  }
  if (below !== undefined && semver.lt(version, below)) {
    return true;
  }
  if (
    excludeRange !== undefined &&
    semver.satisfies(version, excludeRange, { includePrerelease: true })
  ) {
    return true;
  }
  return exclude?.includes(version) ?? false;
};
```

Call site:

```ts
// errorView is available from 1.1.41 on the 1.1 line. The next line branched
// before #3024 forward-merged, so everything from 1.2.0-next.0 up to (not
// including) 1.2.0-next.50 sorts above 1.1.41 without carrying it.
test
  .skipIf(
    crossVersion({
      below: "1.1.41",
      excludeRange: ">=1.2.0-next.0 <1.2.0-next.50",
    }),
  )
```

Checked against real semver rather than prerelease intuition — `includePrerelease: true` is required, and with it the classification is exact:

| version | in range |
| --- | --- |
| `1.1.41`, `1.1.51` | no — keeps running |
| `1.2.0-next.0`, `next.1`, `next.49` | yes — skipped |
| `1.2.0-next.50`, `next.61`, `2.0.0-next.0` | no — keeps running |

No test change needed: there is no unit test for the predicate (it reads the `__FLOW_CROSS_VERSION__` define), and the no-op copy ignores its argument.

**Why the second one:** `exclude: ["1.2.0-next.0"]` is a sample that matches one selector's current output, and it goes stale invisibly — `next` is already on `2.0.0-next.0`, where that entry is dead. `>=1.2.0-next.0 <1.2.0-next.50` states a permanent fact about what was published, so it stays true regardless of what the selector picks next.

**What not to do:** bumping `below` to `"1.2.0-next.50"`. It closes the hole but also skips `1.1.41` … `1.1.51`, which do carry the fix — and `latestOfPreviousLine` picks `1.1.51` at this branch's own version, so that's live coverage lost.

mfal:
Took the second option. Verified the analysis before acting on it, and it holds — with two corrections.

**The hole is exactly 50 versions.** Checked every `1.2.0-next.*` tag rather than the endpoints: `next.0` … `next.49` lack #3024, `next.50` … `next.65` carry it. `>=1.2.0-next.0 <1.2.0-next.50` is precise.

**`latestOfPreviousLine` is `1.1.52`, not `1.1.51`** — a patch shipped since the review. Running the real selector against the real tag list at both `1.2.0-next.59` and today's `1.2.0-next.65`:

```
previous             = 1.2.0-next.64   #3024: has
firstOfLine          = 1.2.0-next.0    #3024: MISSING
latestOfPreviousLine = 1.1.52          #3024: has
```

So `firstOfLine` is indeed the only selector reaching the hole, and the "don't bump `below`" argument is stronger than stated — it would drop `1.1.52`, which is live coverage. I also checked the `offsets` fallback cannot reach in: it only runs when there is no `latestOfPreviousLine`, i.e. on the first line only.

**`next` is on `1.2.0-next.65`, not `2.0.0-next.0`** — no 2.0.0 tag exists yet. Doesn't change the conclusion: the point is that the entry dies silently whenever `firstOfLine` moves off the 1.2 line, not that it already has.

**`includePrerelease: true` is not required for this range.** All eleven rows of your table classify identically with and without it, because both bounds share the version tuple with the tested versions. It *is* required for a range whose bounds do not — `1.1.5-next.3` fails `>=1.1.0 <1.2.0` without it. Kept it, with that as the stated reason rather than "required here".

## Deviation: it does get a test

The rules move to `skipsVersion.ts`, free of the imports that only resolve inside the prepared harness. That was the actual obstacle — `environments.ts` pulls in the old published packages, so a test importing it dies at import. With the split, `unit-dev` (which already globs `e2e/**/*.test.ts`) picks it up. Twelve cases pin the classification across the line boundary; disabling `excludeRange` fails exactly the two in-range ones.

The finding was that the gate is right-by-argument. Correcting the argument leaves it right-by-argument. This makes it right-by-test, which is the same bar `selectTargetVersions.test.ts` already sets for the other half of this machinery.

Verified: eslint, prettier, `test:compile` for both packages, and the 12 new unit tests. The two `ImageCropper` visual scenarios that go red locally are a missing `-webkit-darwin` baseline, not this change — the run generated them and I did not commit them.

The structural pattern you name is filed as #3195.

mfal:
The two red checks are not from this branch. `main` is only the aggregation; the single real failure is `cross-version`, and it is an upstream dependency release.

The iframe harness diverges by one character — a trailing space in the icon `class` attribute:

```
reference (current):                  "... tabler-icon tabler-icon-info-circle "
candidate (1.2.0-next.0 via the host): "... tabler-icon tabler-icon-info-circle"
```

`@tabler/icons-react@3.47.0` shipped on 2026-09-18 and changed `join(" ")` to `filter(Boolean).join(" ")`. The old published package depends on `^3.46.0` and `prepare` resolves it fresh, so the old side lost the trailing space; our current side still renders it, because #3164 inlined the 3.46.0 implementation. Nothing in the repo changed: the same `Button states` comparison against the same `firstOfLine = 1.2.0-next.0` was green in the run on 2026-09-18 at 10:41, and there is no source diff to `Icon` or `Button` between `1.2.0-next.59` (that run's base) and `1.2.0-next.65`.

Fixed at the source in #3202 — one line in `createTablerIcon.ts`, routed to `main` since it is a `fix`. Gating the scenario per version would have been the wrong tool: this is not version-bound, and it would hide a real difference between our output and the package we inlined.

This PR goes green once #3202 forward-merges into `next`.

mfal:
Correction to my previous comment: the diagnosis of the *symptom* held, but the fix I proposed was on the wrong side, and CI showed it. #3202 is closed.

Removing the trailing space from our output flipped `main` red on all three targets. The reason is that the old side's resolution is not deterministic: `prepare` runs a plain `npm install`, so a range like `@tabler/icons-react: ^3.46.0` resolves to whatever is newest at install time — unless the CI cache carries an older install. `main` restored a cache predating 3.47.0 and resolved 3.46.0; this branch missed the cache and got 3.47.0. The harness was reporting cache state.

With resolution pinned, every version published before #3164 legitimately renders the trailing space, so changing our side would diverge from all of them, in every scenario containing an icon.

The real fix is #3209: each old version installs with `--before` its own publish date, so it gets the dependency tree it shipped with. Verified from a cold cache — the iframe harness passes all three targets locally.

This PR goes green once #3209 reaches `next` through the forward-merge.

ins0 (approved):

PR #3196, opened 2026-09-18
test(components): compile the browser tests' base styles from source

### LINKED ISSUE #3194: Browser tests see design tokens from a build artifact, component styles from source
A `.module.scss` change reaches a browser test immediately, but a **design token** added in the same branch does not. The test then renders the component with the custom property missing — and CSS reacts by dropping the whole declaration, not by keeping the previous rule.

Measured while adding `--popover--viewport-padding` in #3193, on a bare `vitest run`:

```
DIV.flow--popover :: maxW=none  varPad=[]  minW=[200px]
```

`max-width: calc(100% - 2 * var(--popover--viewport-padding))` became `max-width: none`, the popover went flush against the viewport edge — the exact layout the change removes. No error, no warning. It reads as "my fix does not work".

## Where the split comes from

`dev/vitest/setupBrowser.ts` loads a build artifact:

```ts
import "../../../stylesheet/dist/styles.css";
```

`packages/stylesheet/build.js` copies `@mittwald/flow-react-components/all.css`, which carries the token layer (`src/styles/index.scss` forwards `@mittwald/flow-design-tokens/css-layered/*`). Component module rules, by contrast, are compiled live by vite from each component's own `.module.scss` import.

So one half of a component's styling is live and the other half is a snapshot. Verified by adding `outline-offset: 7px` to `Popover.module.scss` and running a bare `vitest run` without rebuilding anything:

```
SRCPROBE outlineOffset=7px maxWidth=calc(100% - 32px)
```

The source rule came through; only the token it depends on was stale.

## Why `pnpm nx test:browser` is not the whole answer

The nx target's `^build` chain does rebuild the stylesheet, so CI and a full local run are correct. The trap is the fast iteration loop — `vitest run <file>`, `vitest dev` — where the cost of a full rebuild per change is exactly what one is avoiding. A footgun that only bites during iteration still costs real time, and it costs it while debugging something else.

## Options

1. **Load the token CSS from a live path.** `setupBrowser.ts` imports `@/styles/index.scss` instead of `stylesheet/dist/styles.css`. Vite compiles it, so component rules and the base/token layer come from the same place. The tokens package's `dist/css-layered/*` is still an artifact, but a far cheaper and less frequently changed one, and it is what the source entry already points at.
2. **Generate the token CSS in the vite pipeline**, so the YAML is the input. Removes the artifact entirely; more machinery.
3. **Fail loudly instead.** A setup assertion that a sentinel custom property resolves, erroring with "stylesheet artifact is stale, run `pnpm nx build stylesheet`". Cheapest, does not fix the split, but converts a silent wrong render into a clear message.

Option 1 looks like the right size. Whichever we pick, the check in option 3 is cheap insurance on top.

## Acceptance

- A token added to `packages/design-tokens/src/**` is visible to `vitest run <file>` in `packages/components` without a manual `pnpm nx build stylesheet` — or the run fails with a message that names the missing step.
- `pnpm nx test:browser components` stays green, both the `browser` and `browser-layered` projects. The layered project deliberately loads the *layered* variant, so whatever replaces the import has to keep those two apart.

mfal:
fixes #3194

`dev/vitest/setupBrowser.ts` loaded `stylesheet/dist/styles.css`, a build artifact, while component module rules were compiled live by vite. One half of a component's styling was a snapshot, the other half source — so a design token added in the same branch was missing from `vitest run`, and CSS dropped every declaration referencing it. No error, no warning; it reads as "my fix does not work".

## What changed

Both browser projects import `@/styles/index.scss` — the base/token entry, compiled by vite.

That is **less** than the artifact carried. `stylesheet/dist/styles.css` was a byte copy of `all.css` (`packages/stylesheet/build.js:22`), i.e. the entry *plus* every component module. Component rules now reach the tests only through each component's own module import — which is the point: the copy was the stale half, applied on top of the live modules.

What separates the two projects is the cascade, as before:

- `browser` adds the new `flow-unlayered-styles` PostCSS plugin, which flattens the layers the entry declares — the way `stylesheetVariantsPlugin` derives `all.css` from `all-layered.css`. It is registered there and nowhere else; dev and Storybook import the same entry and keep its layers.
- `browser-layered` keeps them, plus the release build's layer plugin for the component modules.

`stripCascadeLayers` is split so both callers share the walk.

## nx wiring

`test:browser` and `test:browser:dev` carried a project-level `dependsOn` on `stylesheet:build`. That **overrides** `nx.json`'s `targetDefaults` rather than extending it — so the default `^build` never applied, and `stylesheet` reached the rest of the chain only transitively.

Both overrides are deleted, not repointed. `tokens` is already an `implicitDependency`, so the default covers it. Measured task graph for `components:test:browser`:

| | tasks |
| --- | --- |
| `main` | 15 — incl. `stylesheet:build` and the `components:build*` chain |
| repointed at `tokens` (first attempt) | 2 — dropped `icons`, `icons-pro`, `react-tunnel` |
| now | 7 — `tokens`, `icons`, `icons-pro`, `react-tunnel`, no `stylesheet`, no `components:build` |

Thanks @ins0 — the middle row would have broken `pnpm nx test:browser components` on a fresh clone, since those `dist` dirs are gitignored and nothing else builds them.

## The remaining artifact

The token CSS the styles entry forwards stays a build artifact, and stale it fails the same silent way. Issue option 3 proposed a setup assertion for that; it is deliberately **not** here. Every documented entry point is an nx target, and `^build` guarantees a current `dist` — including the fast loop, which has targets too:

```shell
pnpm nx test:browser components -- src/components/Popover --browser.name=webkit
pnpm nx test:browser:dev components
```

Measured: ~2.6 s nx overhead on a single-file run (5.8 s vs 3.1 s bare), paid once at startup in watch mode. A guard would only have covered the bare `vitest` call and its whole message would have been "go through nx" — so `packages/components/AGENTS.md` says that instead, with the reason.

## Verified

- `pnpm nx test:browser components --browser.name=webkit` green across both projects, on `main` merged in.
- The two projects still differ in layering, measured from `document.styleSheets`: `browser` carries no cascade layer, `browser-layered` carries `flow.tokens`, `flow.reset`, `flow.base`.
- A probe token added to `border.yml` with only `pnpm nx build tokens` run: visible to the browser tests, while `stylesheet/dist/styles.css` provably still lacked it.
- Task graphs above computed with `nx run <target> --graph`.
- `test:compile components`, `unit-dev` (165 tests, 4 new for the plugin), eslint, prettier.


ins0 (commented):
The core idea is right, and it fixes a second bug the description doesn't claim: `stylesheet/dist/styles.css` was an artifact of a project that is **not** a dependency of `components`, so it never entered `test:browser`'s cache hash at all. `design-tokens` is one (`implicitDependencies: ["tokens"]`), so token source changes now invalidate properly.

One blocking problem in the nx wiring.

## Blocking

**`packages/components/project.json:142` and `:151` — the `dependsOn` swap drops `icons`, `icons-pro` and `react-tunnel` from the graph, not just `stylesheet` and `components:build`.**

nx merges `targetDefaults` per property, and a project-level `dependsOn` replaces the default outright — it does not concatenate. So `nx.json`'s `test:browser: { dependsOn: ["^build"] }` was already being overridden; on `main` the `stylesheet` edge happened to reach `^build` transitively, and `tokens` does not.

Task graph for `nx run components:test:browser`, computed against current `main`:

| | tasks |
| --- | --- |
| `main` | 15 — incl. `icons:build`, `icons-pro:build`, `react-tunnel:build`, `tokens:build`, `stylesheet:build`, the `components:build*` chain |
| this PR | **2** — `tokens:build`, `components:test:browser` |

`@mittwald/flow-icons` exports only `./dist/esm/components/index.js`, `@mittwald/react-tunnel` only `./dist/index.js`. Both `dist` dirs are gitignored, nothing else builds them, and `src/components/Icon/components/IconSetProvider.browser.test.tsx` imports `@mittwald/flow-icons` directly. After `pnpm clean`, or on a fresh clone, `pnpm nx test:browser components` — the command this PR's own AGENTS.md text tells people to use — cannot resolve the entry. That is the "Common failures" entry about skipping the `^build` chain, reached through the target rather than around it.

**Why CI is green anyway:** the `browser` job's affected set was `["components","remote-react-components"]`. `remote-react-components:test:browser` has no override, inherits `^build`, and built the whole chain; `--parallel=1` then happened to schedule those builds before the test task. Incidental — `components:test:browser` no longer has an edge to them, so the ordering isn't guaranteed at higher parallelism, and a run where only `components` is affected has nothing to piggyback on.

**Fix — delete the override on both targets** rather than repointing it. `tokens` is already in `implicitDependencies`, so the `nx.json` default covers it. Verified graph: `tokens:build`, `icons:build`, `icons-pro:build`, `react-tunnel:build`, `components:test:browser` — 6 tasks. Still drops `stylesheet:build` and `components:build`, which is what the PR was after.

This also restores the premise of the `unit` job's comment in `.github/workflows/test.yml` — "both targets build the same `^build` chain" — which the PR currently falsifies.

## Should fix

**`packages/components/dev/vite/unlayeredStylesPlugin.ts:5` — the docblock names dev and Storybook, where the plugin is not registered.** It is wired only into `vitest.config.ts`'s `browser` project. `vite.config.ts` carries `[unlayeredMarkerPlugin(), autoprefixer()]`, and `.storybook/preview.tsx` imports the same `../src/styles` entry — so in dev and Storybook the base/token rules stay inside `@layer flow.tokens/reset/base`. This PR makes the test project the only place that flattens them, and the comment asserts the opposite of the thing that just became subtle.

## Nits

- **Scope.** The plugin strips every `@layer` from every stylesheet the `browser` project compiles. Safe today — `@layer` in `src/**/*.{scss,css}` is only `styles/index.scss` plus the three `flow.unlayered` markers — but both name and docblock read as if it were scoped to the entry, while the sibling plugins all guard with `isComponentModule`. One sentence, or a guard.
- **Description precision.** "`@/styles/index.scss`, the same entry the release build uses" — the release build's `all.css` is that entry *plus* every component module, and `stylesheet/dist/styles.css` was a byte copy of it (`packages/stylesheet/build.js:22`). Component rules now reach the tests only via each component's own module import. That's the right call — the copy was the stale half, applied on top of the live modules — but the description reads as if nothing changed there.
- **Branch is 411 commits behind `main`.** Merges clean, but CI ran against that base, and the finding above turns on which projects are affected. Rebase and re-run.

## Verified clean

`stripCascadeLayers` split is a faithful extraction. `OnceExit` puts the new plugin after `unlayeredMarkerPlugin`'s `Once` regardless of array position, so the "markers already lifted" premise holds — and the fourth unit test pins it. Extensionless imports in the new test file match every sibling in `dev/vite/`. `tokens:build` genuinely is required (`./css-layered/*` → gitignored `dist/css-layered/*`). The AGENTS.md addition is accurate and in house style. No generated-code, `@flr-generate` contract, or DoD surface is touched — this is test infrastructure, so the component checklist doesn't apply.

mfal:
Thanks — the blocking finding reproduces exactly as described, and the `stylesheet` observation about the cache hash is a better framing of the bug than the one I wrote.

**Blocking — fixed in 45ccf9fc3.** Both overrides deleted rather than repointed. Task graph for `components:test:browser`, via `nx run … --graph`:

| | tasks |
| --- | --- |
| repointed at `tokens` | 2 — `tokens:build`, `components:test:browser` |
| overrides deleted | 7 — `tokens`, `icons` (+`build:icons`), `icons-pro` (+`build:icons`), `react-tunnel` |

Seven rather than your six because the two `build:icons` sub-targets are separate tasks; same set of projects. `stylesheet:build` and the `components:build` chain stay out, so the goal survives. `test:browser:dev` had the same override and is gone too.

I had verified the swap against a worktree that already had every `dist` built, so the missing edges never surfaced — the exact blind spot the "Common failures" entry describes, reached through the target. Noted.

**Should fix — fixed in 318b722bb.** Confirmed: `unlayeredStylesPlugin` appears only in `vitest.config.ts`, and `.storybook/preview.tsx` imports `../src/styles` with the dev config's `[unlayeredMarkerPlugin(), autoprefixer()]`. Docblock now says it runs in the `browser` project and nowhere else, and that dev and Storybook keep the layers.

**Nits.**

- Scope: a sentence, not a guard. `@layer` in `src/**` is `styles/index.scss` plus the three markers, and those are lifted before this plugin runs — the docblock now states that rather than implying a scope it does not have.
- Description: rewritten. It now says `stylesheet/dist/styles.css` was a byte copy of `all.css`, entry *plus* every component module, and that component rules reach the tests through their own module imports only.
- Rebase: merged `main` in instead of rebasing (2ca5756ec) — the branch is public and the PR is already under review. CI is now running against current `main`.

ins0 (approved):

Issue #3195, opened 2026-09-18
Cross-version: a missing per-version gate surfaces late, and on the wrong PR


mfal:
The in-process cross-version harness does not run on pull requests. A PR that introduces a divergence therefore stays green, and the failure appears later — in the scheduled run, or on an unrelated PR that happens to carry the `run-cross-version-tests` label. The author who has to investigate is not the author who caused it.

## What happens today

| Runs | Harness | Trigger |
| --- | --- | --- |
| `test.yml` → `cross-version` | iframe only, ~74s, `nx affected`-gated | every PR |
| `test-visual-scheduled.yml` → `cross-version-inprocess` | broad in-process corpus, per-version matrix | schedule |
| `test-cross-version-label.yml` | both | `run-cross-version-tests` label |

The split is deliberate and documented in `test.yml`: the iframe half checks the attribute-level contract cheaply, the in-process half renders the whole visual corpus per version and costs ~24 minutes of network installs. The consequence is the gap: **the in-process harness is the only one that sees a new component or a changed render structure, and it never sees a PR.**

## Evidence

Two divergences sat ungated on `next` and first surfaced on #3191, a PR about `Action`:

| Change | in 1.1.49 | in 1.2.0-next.0 | Symptom |
| --- | --- | --- | --- |
| `feat(CartesianChart): add Bar …` (2026-08-25) | no | no | `Element type is invalid … got: undefined` in 3 scenarios, both versions |
| `fix(ImageCropper): handle a failed image load` (#3024, 2026-09-15) | yes | no | structural snapshot diff, 1.2.0-next.0 only |

Both were gated in #3191 (`below: barSince`, and `exclude: ["1.2.0-next.0"]`). The `Bar` one had been red on the schedule since late August. Proving that neither came from the PR under review cost a full investigation.

## Two classes, two answers

**1. The component did not exist yet — statically detectable.** A visual scenario destructures the components it uses (`components: { CartesianChart, ChartGrid, Bar, … }`). Whether a component exists in a published version is readable from that version's git tag, no network and no install:

```
git show 1.2.0-next.0:packages/remote-react-components/src/auto-generated/index.ts
```

`Bar` appears there first in `1.2.0-next.6` — exactly the `below:` value the gate needs. So a PR-time check could both **fail with a precise message** and **name the correct gate value**. This class is the one that produces the unhelpful `Element type is invalid` error, and it is the majority of what goes wrong.

[...]

PR #3202, opened 2026-09-21
fix(icons): drop the stray trailing space from the inlined icon class

mfal:
An inlined Tabler icon that gets no `className` renders `class="tabler-icon tabler-icon-info-circle "`. Harmless in HTML — but it currently reds the cross-version iframe harness on every branch, which is how it surfaced.

## Cause

#3164 copied `createReactComponent` from `@tabler/icons-react` at 3.46.0, which built the attribute with a plain `join(" ")`. Upstream 3.47.0 (published 2026-09-18) changed it:

```diff
- [`tabler-icon`, `tabler-icon-${iconName}`, className].join(" ")
+ [`tabler-icon`, `tabler-icon-${iconName}`, className].filter(Boolean).join(" ")
```

Our inlined copy froze the old semantics, so our output now differs from the package it was inlined from.

## Why CI noticed

The cross-version harness renders an old published version through the current host and compares the HTML attribute by attribute. The old package depends on `@tabler/icons-react: ^3.46.0` and `prepare` resolves it fresh on every run — so since 3.47.0 the old side renders *without* the trailing space while our current side still renders it:

```
reference (current):  "... tabler-icon tabler-icon-info-circle "
candidate (1.2.0-next.0 through the current host): "... tabler-icon tabler-icon-info-circle"
```

The same comparison was green on 2026-09-18 at 10:41 and red on 2026-09-21 with no repo change in between — `Button states` / `Button edge cases` against `firstOfLine = 1.2.0-next.0`. It is not a version-bound divergence, so gating it per version would be the wrong tool: it would hide a real difference between our output and upstream's, and it would recur for every scenario containing an icon.

## Verification

Output parity, checked against both upstream implementations:

| `className` | ours after this change | matches 3.47.0 | matches 3.46.0 |
| --- | --- | --- | --- |
| absent | `tabler-icon tabler-icon-info-circle` | yes | no |
| `""` | `tabler-icon tabler-icon-info-circle` | yes | no |
| `"flow--icon"` | `tabler-icon tabler-icon-info-circle flow--icon` | yes | yes |

The empty case is the only behaviour that changes. `icons-pro` renders through FontAwesome and has no such runtime. No test pinned the trailing space — `toHaveClass` and the `.tabler-icon-*` selectors are token-based — and `createTablerIcon.ts` is hand-written (`src/lib/`), not one of the generated artifacts, so nothing needs regenerating.

Icon, Button and Action browser suites pass (37 tests, webkit).

## Note on the file

`packages/icons/src/lib/createTablerIcon.ts` is a frozen copy of third-party code. It will keep drifting from upstream unless something tracks it; worth considering when the next `@tabler/icons-react` bump lands.

related #3195


mfal:
For the record — this fixes the wrong side, and CI proved it.

The premise was that the old published versions render *without* the trailing space, so our inlined copy was the odd one out. That only holds when `prepare` resolves their dependencies fresh, and it does not do so reliably: the CI cache is keyed on `lerna.json` with a `cross-version-<os>-` fallback, so `main` kept an install predating `@tabler/icons-react@3.47.0` and resolved 3.46.0, while the `next`-based branch missed the cache and got 3.47.0.

That is why this branch went red on all three target versions while `main` was green: with the trailing space removed from our side, the cached old side — which still has it — stopped matching. The harness result was tracking cache state, not code.

Once resolution is deterministic the change is actively wrong. Every version published before #3164 resolves a Tabler that emits the trailing space, so dropping it from our output would diverge from *every* old version, in every scenario containing an icon.

The real defect is the non-deterministic resolution. Fixed in #3205: `prepare` installs each old version with `--before` its own publish date. Verified from a cold cache — 1.0.16 and 1.1.0-next.0 pin `@tabler/icons-react` to 3.46.0, and 1.1.52 carries the inlined icons and needs none.

The trailing space stays for now: invisible in HTML, and aligning with upstream is a deliberate output change that belongs with the next Tabler bump, where the version gate can be set in the same commit. Noted on #3195.

PR #3204, opened 2026-09-21
feat(codemods): report the Flow packages' own peer ranges

### LINKED ISSUE #3059: cli(upgrade, list): the Flow packages' own peer ranges are fetched, discarded, and never shown
`upgrade` and `list <revision>` never surface the peer ranges the Flow packages
themselves declare. A consumer crossing a wide range has to open every Flow
package's `package.json` by hand to learn what the new train requires:

| package | peers |
| --- | --- |
| `flow-react-components` | `react`, `react-dom`, `react-hook-form`, `@internationalized/date`, `next`, `@mittwald/react-use-promise` |
| `ext-bridge` | `react`, `react-dom`, `i18next` |
| `flow-icons-pro` | `react`, `@fortawesome/sharp-regular-svg-icons` |
| `remote-react-components`, `remote-react-renderer`, `mstudio-ext-react-components`, `react-tunnel`, `flow-icons` | `react` and friends |
| `remote-core`, `remote-elements` | Flow packages, as **exact** pins |

The package manager does report conflicts — `runInstall` uses
`stdio: "inherit"`, so nothing is swallowed — but in its own terse form, and
only for what it decides to name. Yarn: "react is listed by your project with
version 19.1.0, which doesn't satisfy what `@mittwald/ext-bridge` and other
dependencies request (^19.2.0)". "other dependencies" is never enumerated.

## The data is already in hand

`fetchVersions` requests the abbreviated packument
(`application/vnd.npm.install-v1+json`), which carries `peerDependencies` and
`peerDependenciesMeta` per version. `src/resolve/registry.ts` reduces the whole
document to `Object.keys(packument.versions)` and discards the rest. A summary
therefore costs **no extra request** — and works in `--dry`, on an install
failure, and when there is no bump at all.

## Proposal — print, don't judge

- Keep the target version's peer ranges in `fetchVersions` / `RegistryVersions`.
- Surface them from `resolveRange`, which `upgrade` and `list <revision>`
  already share — so `list` answers "which peers will I need?" *before*
  anything is written, which is where the answer is worth most.
- Print a Flow-authored summary grouped by peer (`react ^19.2.0 — required by
  …`), so the list the manager truncates is complete.

**Non-goal: deciding whether the consumer satisfies them.** That is a second
resolver beside the package manager's, it can drift, and it can be wrong. The
manager owns satisfaction. Ranges printed as authored cannot be.

Worth calling out separately: Flow's internal peers publish as **exact** pins,
not ranges — `@mittwald/flow-remote-core@1.1.52` peers on
`@mittwald/flow-react-components: "1.1.52"`. With the publish holes of #2887
(Lerna publishes only changed packages), an exact peer pin can name a version
its peer never published. `resolveRange` already resolves against the
intersection of the **declared** set; nothing shows the consumer these pins.

---

*Rewritten from the original report, which described an asymmetry between the
install-failure and install-success branches. There is none: neither branch
reports peers — the failure branch's message is install recovery (`package.json`
bumped, `node_modules` stale, how to undo). The `peerDependencies` line
`upgrade` does print concerns the **consumer's own** peer field on Flow packages
and is deliberate (README: "reported and left alone") — the opposite direction
from this issue.*

ins0:
Closes #3059.

`upgrade` and `list <revision>` never surfaced what the Flow packages themselves peer on. Answering "which Flow package needs which peer range" meant opening every Flow `package.json` by hand. The package manager reports conflicts, but only the ones it decides to name — Yarn's "`@mittwald/ext-bridge` and other dependencies" never enumerates the others.

## Before / after

`upgrade major` on a project sitting on `0.2.0-alpha.584`:

**Before**

```
Upgrading Flow from 0.2.0-alpha.584 to 1.1.48
  @mittwald/flow-react-components → 1.1.48
  @mittwald/ext-bridge → 1.1.48
Installed with pnpm — pnpm install --no-frozen-lockfile
  renamed-css-export: 1 file(s) changed
```

**After** (real registry data at 1.2.4)

```
Upgrading Flow from 0.2.0-alpha.584 to 1.2.4
  @mittwald/flow-react-components → 1.2.4
  @mittwald/ext-bridge → 1.2.4

Peer requirements at 1.2.4 — what the Flow packages you declare ask of your
project:

  @internationalized/date ^3.12.2 — optional
    @mittwald/flow-react-components
  @mittwald/react-use-promise ^4.2.2 — required
    @mittwald/flow-react-components
  i18next ^26.0.0 — optional
    @mittwald/ext-bridge
  next ^16.2.3 — optional
    @mittwald/flow-react-components
  react ^19.2.0 — required
    required by @mittwald/flow-react-components
    optional for @mittwald/ext-bridge
  react-dom ^19.2.0 — required
    required by @mittwald/flow-react-components
    optional for @mittwald/ext-bridge
  react-hook-form ^7.65.0 — optional
    @mittwald/flow-react-components

Ranges as Flow declares them, not a check of what you have installed — your
package manager decides that during the install. An optional peer need not be
installed at all; it only has to match the range if it is.

Installed with pnpm — pnpm install --no-frozen-lockfile
```

Each peer carries its status, because that is the reader's question: **must I
install this at all?** One package requiring it settles that however many
others call it optional, so `react` reads `required` and the two sides are
broken out underneath. Optionality is per package and comes from
`peerDependenciesMeta`, which the same packument carries.

A Flow-internal pin the target does not satisfy gets its own section. Flow publishes those as exact versions, so a divergent one usually means the pinned version was never published (#2887):

```
1 Flow-internal peer pin(s) do not accept 1.1.48. Flow publishes these as exact
pins, so this normally means the pinned version was never published (#2887) — …

  @mittwald/ext-bridge 1.1.44
    @mittwald/flow-remote-react-components
```

## How

**No extra request.** `fetchVersions` already requests the abbreviated packument, which carries `peerDependencies` per version; `registry.ts` reduced the whole document to `Object.keys(packument.versions)` and discarded the rest.

- `resolve/peers.ts` — `collectPeers` groups by peer **and** range, so two Flow packages disagreeing about the same peer stay two entries instead of collapsing into one.
- Resolved in `resolveRange`, which `upgrade` and `list` already share, so `list <revision>` answers it before anything is written.
- `upgrade` prints it **before** the install in every branch (bump, `--dry`, already-on-target). A failed install returns early, and that is exactly when the reader wants the ranges on screen — there is a test asserting the order, not just the presence.
- `list --json` carries `peers: { external, flowPins }` — each entry `{ peer, range, requiredBy, optionalFor }` — and `null` for the offline whole-catalogue form.

**Print, don't judge.** Ranges appear as authored; nothing here reads `node_modules` or decides whether the project satisfies them. That would be a second resolver beside the package manager's, and it could only drift from the one that actually gates. Hence the closing line in the output — without it the list reads as a check that passed.

## Incidental

`wrap`/`painter`/`stripAnsi` move from `list.ts` to `cli/text.ts` verbatim. `list.ts` renders the peer section too, so leaving them where they were would make `list.ts` and `peers.ts` circular.

## Verification

- `pnpm nx test:unit codemods` — 390 pass (25 new)
- `pnpm nx test:compile codemods` — clean
- `pnpm lint` — exit 0
- `pnpm nx build codemods` — no generated drift

No consumer-facing breaking change, so no migration catalogue entry: this only adds output.


ins0:
## 🚀 Preview Deployment

Preview environments are ready:

| Type | URL |
|------|-----|
| docs | [pr-3204.docs.review.flow-components.de](https://pr-3204.docs.review.flow-components.de) |
| storybook | [pr-3204.storybook.review.flow-components.de](https://pr-3204.storybook.review.flow-components.de) |

Images:
- docs: `ghcr.io/mittwald/flow/docs:pr-3204`
- storybook: `ghcr.io/mittwald/flow/storybook:pr-3204`

Lisa18289:
## Review

Checked out locally against `next`: `test:unit codemods` 384 pass, `test:compile` clean, eslint + prettier green, PR CI fully green. The "no extra request" claim holds — the abbreviated packument does carry `peerDependencies`.

The design decisions carry: resolving in `resolveRange` so `list` and `upgrade` cannot drift, printing before the install in all three branches with a test asserting the *order* rather than mere presence, `peers: null` instead of omitted in the JSON, grouping by peer **and** range so a disagreement survives. Deliberately not re-implementing the package manager's resolution is the right call.

Two things before merge.

### 1. Optional peers are presented as required

I ran `collectPeers` + `renderPeers` against the real packuments of `@mittwald/flow-react-components@1.2.2` and `@mittwald/ext-bridge@1.2.2` — the common consumer setup. Actual output:

```
  @internationalized/date ^3.12.2     <- optional
  @mittwald/react-use-promise ^4.2.2  <- required
  i18next ^26.0.0                     <- optional
  next ^16.2.3                        <- optional
  react ^19.2.0                       <- required (frc), optional (ext-bridge)
  react-dom ^19.2.0                   <- required (frc), optional (ext-bridge)
  react-hook-form ^7.65.0             <- optional
```

Five of seven lines are wholly or partly optional and nothing says so. Under "what the Flow packages you declare ask of your project", a Vite-based extension project reads that it needs Next 16. The "After" example in the PR description shows exactly this. This is the question the feature exists to answer, and for most lines it answers it wrongly.

The data is free — same document, same request:

```
keys: name,version,dependencies,devDependencies,peerDependencies,dist,engines,peerDependenciesMeta
peerDependenciesMeta: {"react":{"optional":true},"i18next":{"optional":true},"react-dom":{"optional":true}}
```

Read `peerDependenciesMeta` in `PackumentVersion`, carry it through `RegistryVersions` into `PeerRequirement`, mark it in the renderer, and expose it in `list --json` — otherwise a JSON consumer cannot recover it either. Note optionality is **per package**, not per range: `react` is required by `flow-react-components` and optional for `ext-bridge`, and the `[peer, range]` key collapses both into one line today.

### 2. `flowPins`: a reproducible false positive, and otherwise unreachable

`expectedFlowPin` calls `satisfies(target, range)` without `includePrerelease`. A prerelease version satisfies no non-prerelease range in semver — not even `*`. Real registry data, `@mittwald/flow-remote-core` at target `0.2.0-alpha.103`:

```
1 Flow-internal peer pin(s) do not accept 0.2.0-alpha.103. Flow publishes these
as exact pins, so this normally means the pinned version was never published
(#2887) - ...

  @mittwald/ext-bridge *
```

`*` accepts everything, and a publish hole is reported anyway. This hits every prerelease target, and those are live dist-tags: `next` -> `1.3.0-next.3`, `experimental` -> `0.2.0-experimental.776`. One-line fix: `satisfies(target, range, { includePrerelease: true })`.

That fix raises the bigger question. I evaluated every published version of the five remote packages (~6200 Flow-internal peer entries): historically the pin is `*`, and today pnpm writes the package's own version exactly at publish time, so the pin always equals `target` by construction. With the prerelease fix, `flowPins` is empty across the entire publish history — the false positive is its only observable effect today. Consider dropping the section, or say in the PR why it stays.

### Smaller

- **README "This list is complete"** — complete only over the *directly declared* Flow packages. A project declaring just `flow-react-components` never sees what `flow-remote-core` peers on. Half a sentence, in the spirit of the rest.
- **`localeCompare` for sorting** — ICU-dependent, notably for hyphenated names. A plain `<` is deterministic.
- **The branch is 422 commits behind `next`** (merge base: `chore(release): bump version to 1.1.48`), which is why the example output cites 1.1.48 while latest is 1.2.2. Rebase before merge.

The irony of #1: the discipline this PR sets for itself — report what is declared, judge nothing — is what dropping `peerDependenciesMeta` breaks. Omitting it *is* a judgement, and the wrong one.

ins0:
Thanks — both findings were real. Fixed in 88a880a, rebased onto `next` (the branch was 435 behind by then). One conclusion I have to push back on, with numbers.

## 1. `peerDependenciesMeta` — accepted, and it was worse than the sample showed

Verified against the live packuments. At 1.2.4 **all three** of `ext-bridge`'s peers are optional, plus `next`, `react-hook-form`, `@internationalized/date` and `flow-icons-pro` on `flow-react-components`.

The two fields are now joined at the registry boundary into one `PeerDeclaration { range, optional }`, so nothing downstream can read a range without its flag. Optionality is recorded per package as `requiredBy` / `optionalFor` inside the group rather than splitting it, and `list --json` carries both — with a test asserting the optional peer survives into the JSON, since a consumer reading only `requiredBy` would reproduce the human-side bug.

The status also moved onto the peer line, in colour, because the reader's question is "must I install this at all?" — one package requiring it settles that however many others call it optional. Real output at 1.2.4:

```
  next ^16.2.3 — optional
    @mittwald/flow-react-components
  react ^19.2.0 — required
    required by @mittwald/flow-react-components
    optional for @mittwald/ext-bridge
```

The two sides are broken out only when they disagree; otherwise the status carries it alone.

## 2. Prerelease — fix accepted, the conclusion does not hold

`includePrerelease: true` is in. Confirmed: `satisfies("0.2.0-alpha.103", "*")` is `false` without it, and `next` / `experimental` are prerelease dist-tags, so this was a live path.

But **`flowPins` is not empty once fixed.** I ran the same sweep over the five remote packages' full published history — 6210 Flow-internal peer entries, each version evaluated at its own version as target:

| | entries |
|---|---|
| flagged today | 4581 |
| of those, `*` — the false positive you found | 2328 |
| **still flagged with `includePrerelease`** | **2253** |

The remainder are not `*` and not the package's own version — they are cross-version pins:

```
@mittwald/flow-remote-core@0.2.0-alpha.700
  peers -> { "@mittwald/ext-bridge": "0.2.0-alpha.699", ... }
```

Upgrading every declared package to one target leaves that pin unsatisfied. That is exactly the shape the section exists to surface, so it stays. The false positive was one shape class (`*`), not the feature.

Both sides now have a test: a `*` pin at a prerelease target reports nothing, a cross-version pin at a prerelease target still reports.

I suspect the difference is the target: evaluated against the *latest* version the old pins are all historical noise, but `upgrade` moves every declared package to one target, which is the case that makes them bite.

## Smaller — all taken

- **README "complete"** — now says complete over the packages the manifest *declares*, with the `flow-remote-core` example.
- **`localeCompare`** — replaced with a plain string compare, with the ICU reason in the comment.
- **Rebase** — done, onto current `next`. The examples in the description now cite 1.2.4.

Not addressed, deliberately: nothing here reads `node_modules` or decides satisfaction. That stays the package manager's call.

Lisa18289:
## Review

Rebase and both fixes hold up locally — 390 unit tests green, `test:compile` and lint clean, CI green. Joining `peerDependenciesMeta` into `PeerDeclaration` at the registry boundary is the right place for it: nothing downstream can read a range without its flag any more.

Three things.

### 1. A package can show up twice

`findFlowDependencies` returns one entry per manifest *field*. A Flow package declared in both `peerDependencies` and `devDependencies` — the usual setup for a library that wraps Flow — therefore arrives twice in `fetched` and lands twice in `requiredBy`. Real CLI output against such a manifest:

```
  react ^19.2.0 — required
    @mittwald/flow-react-components, @mittwald/flow-react-components
```

Every line is affected, `list --json` included. Deduplicating the names before `fetchAllVersions` fixes this and the duplicate registry request in one go. No test covers the case.

### 2. The `flowPins` explanation is wrong for almost everything it flags

The section itself earns its place — an exact pin that does not accept the target really does conflict at install time. The sentence under it does not: *"this normally means the pinned version was never published (#2887)"*.

Your sweep and mine differ a little in package set, but the count is not the point — the split inside it is. Of the cross-version pins I see (2059), **2018 point at a version that is published**. Only 41 are actual publish holes, 14 of them on an optional peer. Your own example is in the 2018: `@mittwald/ext-bridge@0.2.0-alpha.699` exists. So does the only hit reachable today, `flow-icons-pro 0.2.0-alpha.775` at the `experimental` target; at `1.2.4` and `1.3.0-next.5` the section is empty.

The ordinary cause is the other one: Lerna publishes only what changed, so the pin names a neighbouring release. That is a different situation with a different thing for the reader to do. Worth saying that instead — the publish hole is the one-in-fifty case.

### 3. The README section did not move with the fix

`88a880a` touched only the `--json` line. The prose is still the pre-fix version: the sample output shows `react ^19.2.0` without the `— required` status, optional peers are not mentioned at all, "This list is complete" is still unqualified (your reply says it now names the `flow-remote-core` case — that is not in the diff), the Flow-internal paragraph repeats the wrong cause from #2, and the example still cites 1.1.48.

Only #1 is a bug; the rest is text.

ins0:
All three confirmed, fixed in bacb596. One of them was a false claim on my side — that first.

## 3. The README, and what I told you about it

You are right, and my previous reply asserted otherwise. Only the `--json` line landed. The prose edit was a scripted string replacement that matched nothing after `pnpm format` had rewrapped those paragraphs, and the script reported success regardless — so I wrote "it now names the `flow-remote-core` case" while the diff said nothing of the sort. Apologies for making you check that twice.

The section now carries the status markers, per-package optionality via `peerDependenciesMeta`, "the Flow packages your `package.json` **declares**" in place of the unqualified "complete", the `flow-remote-core` example, the corrected pin cause, and 1.2.4 throughout.

## 1. Two-field dependency — confirmed, fixed

Reproduced exactly as you described:

```
findFlowDependencies -> [{"field":"devDependencies",...},{"field":"peerDependencies",...}]
collectPeers        -> requiredBy: ["@mittwald/flow-react-components",
                                    "@mittwald/flow-react-components"]
```

Deduplicated before `fetchAllVersions`, so the duplicate request goes with it. `dependencies` keeps both entries on purpose — the two fields are still reported separately, one rewritten and one left alone. Test in `range.test.ts` asserts one request and one name; I checked it fails without the fix (`expected [ …(2) ] to deeply equal [ … ]`) rather than passing either way.

## 2. The cause was wrong — confirmed

My sweep agrees with yours where it counts. Of the 2253 cross-version pins:

| | entries |
|---|---|
| pinned version **is** published | 2207 (98.0%) |
| pinned version genuinely missing (#2887) | 46 |

So "normally means the pinned version was never published" was wrong for 98% of what it flags. Same ratio as your 2018/41, different package set. The message now leads with the real cause:

> N Flow-internal peer pin(s) do not accept 1.2.4. Flow publishes these as exact pins and releases only the packages that changed, so a pin usually names a neighbouring release rather than the version you are installing — your package manager will report the conflict. Rarely the pinned version was never published at all (#2887).

The code comment on `flowPins` carries the same split, so the next reader does not have to re-measure it.

391 unit tests green, `test:compile` and lint clean.

Lisa18289 (approved):

Issue #3030, opened 2026-08-31
prepare-release: capture the release-notes screenshots automatically


mfal:
`/prepare-release` should capture the release-notes screenshots itself instead of
leaving a placeholder for the maintainer.

## Current state

`.claude/templates/release-notes.md` instructs the opposite:

> Screenshots/GIFs: leave the placeholder for the maintainer — do NOT fabricate
> images.

The rule against *fabricating* images is right. Leaving the placeholder is the
part that does not hold up: every minor/major promotion PR ships an unfilled
`<!-- ![caption](url) -->`, and filling it becomes manual work at the exact
moment the maintainer wants to merge. In #3029 (the `1.1.0` promotion) that was
one placeholder in the `Rating` section.

A screenshot taken from a Storybook story is not fabricated — it is the real
component rendered from the real code on the release branch. That is exactly
what the command should produce.

## It is mechanizable — verified while preparing #3029

The figure below was produced from the `release/1.1.0` branch in one headless
pass, no hand-editing:

1. `pnpm nx dev components` (Storybook), then read `/index.json` — it lists
   every story id, e.g. `form-controls-rating--with-segments`.
2. Drive variants through the story URL:
   `/iframe.html?id=form-controls-rating--default&args=maxValue:10;fill:single&viewMode=story`.
3. Stack the variants as same-origin iframes in one page, one monospace caption
   per panel naming the prop it demonstrates, and screenshot that page at
   `deviceScaleFactor: 2`.

## Traps to encode in the command

Each of these cost time and each is verified, not theoretical:

- **Playwright resolves only from the repo root.** A script in a scratch
  directory dies with `ERR_MODULE_NOT_FOUND: Cannot find package 'playwright'`.
  `node --input-type=module -e "$(cat script.mjs)"` with the repo root as cwd
  works.
- **Storybook filters URL args to declared args.** `maxValue` and `fill` are in
  the story's `argTypes` and applied; `defaultValue` is not, so
  `args=defaultValue:7` was dropped **silently** — the screenshot still renders,
  just not the state that was asked for. Assert the rendered DOM (e.g. the
  `input[type="radio"]` count) instead of trusting the URL.
- **Size an iframe by the content's `.bottom`, not its `.height`.** The rect is
  relative to the iframe viewport, so `.height` alone omits Storybook's body
  `padding-top` and clips that many pixels off the bottom of the story. Use
[...]

PR #3211, opened 2026-09-21
chore(prepare-release): stop amending a merge commit that is usually not there

mfal:
Step 10 of `/prepare-release` merged `<to>` into the release branch and then
folded the changelog restore into that merge with `git commit --amend --no-edit`.

The forward-merge cascade leaves `<to>` a strict **ancestor** of `<from>` in the
healthy steady state, so the merge prints "Already up to date." and writes no
commit. `--amend` then rewrites whatever `HEAD` happens to be — `<from>`'s own
`chore(release): bump version to x.y.z-next.N`, a commit already published on
`origin/<from>`.

Hit while preparing the `1.2.0` promotion (#3210): `origin/next..origin/main`
was empty and no merge commit was produced.

## Change

- Commit the changelog restore on its own, guarded by `git diff --cached --quiet`
  so it is a no-op when the changelogs already match.
- Say in the recipe that the merge is **expected** to do nothing, and that no
  step may assume a fresh merge commit exists.
- Keep the merge for the case where `<to>` does carry something `<from>` never
  received — that is when it makes the PR mergeable, and the `.gitattributes`
  driver note still applies there.
- Correct the closing "You do NOT" section, which claimed a single commit.

The changelog restore itself is unchanged in intent: it is about content
(`<from>`'s changelog carries the prerelease line), not about the merge.

related #3030 — which recorded this defect while it was still only observed


mfal:
Superseded by #3227, which carries the same fix as part of resolving #3030.

The `--amend` is not merely wrong there any more but actively destructive: #3227 adds a figure commit right before it, so amending would rewrite that commit's SHA and break every image URL in the release notes. The restore therefore had to move into the same change.

One deliberate difference: the commit subject is `chore(sync): take the stable changelogs from <to>`, not `chore(release): restore …` — #3030 records that choice, to keep the `chore(release):` prefix (the `publish.yml` skip-guard's pattern) off any commit that is not the graduation.

PR #3224, opened 2026-09-21
fix(CoachMark): play the closing animation instead of vanishing

mfal:
A coach mark disappeared in the frame it was closed, while a contextual help fades and slides back toward its trigger.

## Cause

Both render `Popover`, but `CoachMark` sets `modality="non-modal"` and so goes through `NonModalPopoverContent` — a hand-rolled popover that skips react-aria's `usePopover`. It unmounted on `if (!isOpen) return null`, so the element was gone in the render that closed it and the `&[data-exiting]` rule `Popover.module.scss` already defines never matched. `ContextualHelp` takes the modal path, where `Aria.Popover` keeps the element mounted for the exit.

## Fix

Mirror what `Aria.Popover` does, with the same react-aria hooks:

- **Split the component.** The outer half keeps the ref, calls `useExitAnimation` and decides mounting (`!isOpen && !isExiting → null`); the inner half holds the positioning and remounts on every open. The animation cannot live on an element the closing render unmounts, and the entry animation needs that remount to reset.
- **Render `data-exiting`**, which is what the existing reverse-slide rule selects on.
- **Replace the hand-rolled `isEntering`** (`useState` + `onAnimationEnd`) with `useEnterAnimation(ref, isPositioned)`. This fixes a second, quieter bug: the entry animation started while the popover was still `visibility: hidden` waiting for its placement, so it often finished unseen. Keeping it would also have collided with the exit — it reset `isEntering` to `true` on close, and the element would have carried `data-entering` and `data-exiting` at once.

No prop changes, so nothing regenerates.

## Verification

Measured in Storybook (real WebKit, animations on), paused at 220 ms of the 300 ms exit: `flow--popover--popover-slide`, `direction: reverse`, `opacity: 0.10`, `translateY(-7.16px)` — fading and sliding back toward the anchor, same as `ContextualHelp`.

Green locally: `test:compile`, the full `browser (webkit)` project, the layered project for the popover components, `pnpm lint`. Visual suite green in CI via `run-visual-tests`.

A browser test covering the exit was written and then dropped on review request. It required `commands.setReducedMotion("no-preference")`, because the suite forces `reduce` in `dev/vitest/setupBrowser.ts` and that zeroes every animation — without the opt-out such a test passes on the broken code as well. So the fix currently has no regression gate: a future change that unmounts the popover on close again goes unnoticed.


packages/components/src/components/CoachMark/CoachMark.browser.test.tsx
```diff
@@ -507,3 +507,33 @@ test("Without a width the coach mark keeps its own cap", async () => {
 
   expect(content?.getBoundingClientRect().width).toBe(360);
 });
+
+test("A dismissed coach mark animates out instead of disappearing", async () => {
```
mfal: Remove this test

Lisa18289 (approved):

PR #3233, opened 2026-09-22
test(CoachMark): gate the cross-version scenarios per version

Lisa18289:
The scheduled in-process cross-version harness is red on `main` with six failures against `1.2.0-next.0` and `1.1.53`. Two of them are CoachMark, and those are what this fixes.

CoachMark landed in 1.2.0-next.62, so both scenarios resolve it to `undefined` on either tested old version and the host dies with "Element type is invalid". Not a regression — the component is newer than the versions it is compared against. `crossVersion` is a no-op in the normal visual suite, so Local, Remote and their baselines are unaffected.

The other four failures (`Bar` ×3, `ImageCropper` `errorView`) are already fixed on `next` in #3204 and reach `main` with the next release merge, so they are deliberately not duplicated here.


mfal:
Two points from reviewing this. The first changes one line here, the second is separate.

### `coachMarkSince` should be `1.2.0`

Behaviourally identical, but the better statement.

All three currently selected targets — `1.2.4`, `1.2.0-next.0`, `1.1.53` — are treated the same by either bound. They diverge only on `1.2.0-next.62` … `next.67`, and `selectTargetVersions` never picks those: `firstOfLine` takes the *earliest* version of the line (`next.0`), `previous` the nearest below current (`1.2.4`).

So it is about what the constant says:

- `1.2.0` is the contract — the first version in which a consumer can have CoachMark at all. `1.2.0-next.62` names an internal snapshot nobody could install.
- Rounding up is safe: it can never under-skip, only over-skip, and only across versions that are not targets anyway.
- Precedent exists: `Rating.browser.test.tsx` uses `const ratingSegmentSince = "1.1.0"` with "part of the next minor" — the stable minor, not the prerelease it landed in.
- CONTRIBUTE.md's "determine the boundary rather than guessing" is about *structural* divergences, where you have to bisect. "The component did not exist yet" is in the history; nothing to guess.

The precise fact belongs in the comment, not in the constant:

```ts
// CoachMark landed in 1.2.0-next.62 and shipped in 1.2.0.
const coachMarkSince = "1.2.0";
```

### Should prereleases be cross-version targets at all?

"Round to the stable minor" only holds cleanly once no prerelease is a target. What `selectTargetVersions.ts` resolves against the real published list today:

```
main 1.2.5          → previous=1.2.4, firstOfLine=1.2.0-next.0, latestOfPreviousLine=1.1.53
next 1.3.0-next.8   → previous=1.3.0-next.7, firstOfLine=1.3.0-next.0, latestOfPreviousLine=1.2.5
```

- A prerelease carries no compatibility promise. Deprecate-don't-break covers released API; a prerelease is by definition where things still move.
- No consumer is there. `1.2.0-next.0` was reachable only via the `next` tag, in a window that closed when 1.2.0 shipped. The tag has long since moved to `1.3.0-next.8`.
- `firstOfLine` does not mean what it picks. The intent is "oldest version of the current line still in the field" = `1.2.0`. `lineOf()` strips the prerelease tag, so 68 `1.2.0-next.*` queue up in front and one of them wins.
- On `next` it is worse: two of three targets come from the **same in-flight cycle**. A feature landing in next.5 that changes host output breaks the next.0 comparison immediately — the cycle is diffed against itself.

The cost is a standing tax: every feature landing mid-cycle needs a `crossVersion({ below })` gate for a version nobody runs. This PR is one instance.

Caveat — do not filter prereleases globally. Roughly 1100 of the 1145 published versions are `0.2.0-alpha.*`, and before 1.0 that was the only thing extension developers could build against; the offset fallback exists for exactly that. The rule would be: prereleases only when a line has no stable release. Restricting the candidates for the semver categories to stable yields:

```
main 1.2.5          → previous=1.2.4, firstOfLine=1.2.0, latestOfPreviousLine=1.1.53
next 1.3.0-next.8   → previous=1.2.5, offset-10=1.1.50
```

One edge there: on `next`, `latestOfPreviousLine` collapses into `previous`, `push()` dedupes it away, and `hasLatestOfPreviousLine` goes false → offset fallback. Fine in substance, but that check should be about coverage, not about the array entry.

Also stale: CONTRIBUTE.md still describes the offset fallback as "for prerelease-only histories (the current state)".

Either way this PR stands on its own — the gate is needed for `1.1.53` regardless, since CoachMark shipped in 1.2.0.

Lisa18289:
Agreed on the direction — a prerelease from a cycle still in flight is not an "old published version" in the sense this suite means. I ran the proposal against the real published list to check it.

`selectCrossVersionTargetVersions` with the candidates restricted to stable:

```
current 1.2.5         today        previous=1.2.4, firstOfLine=1.2.0-next.0, latestOfPreviousLine=1.1.53
                      stable-only  previous=1.2.4, firstOfLine=1.2.0,        latestOfPreviousLine=1.1.53

current 1.3.0-next.8  today        previous=1.3.0-next.7, firstOfLine=1.3.0-next.0, latestOfPreviousLine=1.2.5
                      stable-only  previous=1.2.5, offset-10=1.1.50
```

Both match what you wrote.

### The caveat and the example contradict each other

"Do not filter prereleases globally" is aimed at the alpha era — but the `next` output you give, `previous=1.2.5, offset-10=1.1.50`, is what a *globally* stable-only list produces, and it carries no alpha target at all. It cannot: only 77 of the 1145 published versions are stable, so `offset-100` and `offset-200` index below zero and are skipped. Only `offset-10` survives, and the fallback drops from four targets to two.

That is live, not hypothetical — `2.0.0-next.0` is published:

```
current 2.0.0-next.0  today        previous=1.3.0-next.8, offset-10=1.2.5, offset-100=1.1.37, offset-200=0.2.0-alpha.1026
                      stable-only  previous=1.2.5, offset-10=1.1.50
```

So the rule has to say which list each path gets: restrict the three semver categories to stable, leave the offset fallback on the full list. Then 1.2.5 and 1.3.0-next.8 come out as you wrote, and 2.0.0-next.0 keeps its alpha reach.

(Minor: 977 of the 1145 are `0.2.0-alpha.*`, not ~1100. Doesn't change the argument.)

### `hasLatestOfPreviousLine`

Confirmed. On `next`, `latestOfPreviousLine` resolves to 1.2.5, which equals `previous`; `push()` dedupes it, the flag goes false, the fallback runs — and re-adds 1.2.5 anyway. Right answer, reached by accident. Agreed the check should ask whether a previous line is covered, not whether the array grew.

### What the change gives up

With no prerelease target, an in-flight cycle gets no coverage of its own line until it ships. I read that as the point rather than the cost — a host-output change inside one cycle breaks compatibility with nothing published — but it is the one thing today's `firstOfLine` does that the proposal stops doing, so it should be a decision rather than a side effect.

Lisa18289:
Housekeeping on the green cross-version comment above: **it does not cover the current head.**

It was posted for run [35706789817](https://github.com/mittwald/flow/actions/runs/35706789817) on `bf69b421d` — a commit that has since been force-pushed away. That version of the branch also carried cherry-picks of the `Bar` and `ImageCropper` gates from #3204, which is why both harnesses were green. The comment has a fixed marker and is updated in place, so it stayed put while the branch moved.

What actually ran on `02f675d0a`:

- **Iframe harness** — `cross-version` job in `Run tests`: ✅ passed.
- **In-process harness** — did not run. `Run Cross-Version Tests` is `skipped`; I removed the `run-cross-version-tests` label on purpose.

The label came off because with only the CoachMark gate the in-process harness still fails `CartesianChart with Bars` ×3 and `ImageCropper custom error view` — the four failures this PR deliberately leaves to #3204's forward-merge. A labelled run would go red on things the PR does not claim to fix, which is noisier than not running it.

Verified locally instead, full run against all three targets:

```
previous = 1.2.3              CoachMark ×2 run, pass
firstOfLine = 1.2.0-next.0    CoachMark ×2 skipped; 4 failed (Bar ×3, ImageCropper ×1)
latestOfPreviousLine = 1.1.53 CoachMark ×2 skipped; 3 failed (Bar ×3)
```

Identical before and after rounding the bound to `1.2.0`.

mfal (approved):

PR #3234, opened 2026-09-22
fix(PasswordCreationField): make the field usable with a screen reader

### LINKED ISSUE #2498: PasswordCreationField is not optimized for screen reader
The **PasswordCreationField** is not optimized for screen reader users. Important information is not announced clearly, which makes it hard to understand the field, the password rules, and the validation status.

1. The info button does not clearly explain that it contains password requirements. 
2. After that, the generate button is focused without enough context about what will be generated.
3. When entering a password, the screen reader only says that the password is invalid, but does not explain which rule is not met. 
4. The show and hide button does not inform the user when the password visibility changes.
5. The password rules in the popover are also not announced as one clear message. 
There is also a separate issue related to the popover: the [popover tip should not be announced by screen readers](https://github.com/mittwald/flow/issues/2499).

ins0:
## What & why

Closes #2498.

The field reported itself as invalid without ever saying why, and three of its controls carried no state a screen reader could read. Verified against Chrome's accessibility tree (CDP `Accessibility.getFullAXTree`) before and after.

**The field error was never wired to the input.** `useFieldComponent` only saw the `isInvalid` **prop**, not the state derived from the policy validation, and its `fieldProps` were computed and then dropped instead of spread onto the `TextField` (`TextField` itself spreads them). So `aria-describedby` was `null` while invalid — the success message was described, the failing rule was not. A screen reader announced "invalid" and nothing else. Issue point 3.

**A description is only read on focus,** and the field keeps focus while the password is typed. The validation message is therefore also announced politely through `@react-aria/live-announcer`, once per state change — the same pattern `Action` and `List` already use. Issue point 3.

**The generate button** announced "Generate" with nothing saying what is generated; it now carries a localized `aria-label`. Issue point 2.

**Toggling the visibility** swapped the input type and the button label with no announcement. Showing or hiding the password is announced now. Issue point 4. (Label swap + live region, as GOV.UK's password input does it — not `aria-pressed`, which would fight the changing label for the accessible name.)

**The rules in the contextual help** were loose lines of text. They are a `<ul>` labelled by its heading now, and each rule states whether it is met — "Not fulfilled: At least 12 characters" instead of a warning icon alone. The icon became decorative, so its generic "Status «warning»" no longer doubles every rule. Issue point 5.

Issue point 1 (the info button's label) was already fixed by the `subject` prop in #2605.

Two suggestions from the issue were deliberately not taken:
- **`role="group"` around the field** — the field already has a single labelled control; a group role would add "group, Password" in front of every announcement without adding information.
- **`aria-describedby` on the generate button pointing at the field label** — the label element contains the buttons, so the description would repeat both button labels back. An `aria-label` on the button is the smaller, correct fix.

## Rendering

Unchanged. The visual suite passes against the existing baselines (`webkit-darwin`, all 6 scenarios), and a Storybook A/B screenshot of the open rules popover is byte-identical before and after.

## Tests

Five browser tests in `PasswordCreationField.browser.test.tsx`, each confirmed to fail without its fix (12 → 13 tests; 5 fail when the source changes are stashed).

## Out of scope — a separate `Label` defect found on the way

Every form field that puts a button in its `Label` leaks that button's name into the **input's** accessible name, because the input's `aria-labelledby` points at the `<label>` element and the accessible-name computation walks its whole subtree. Chrome's AX tree for this field:

```
textbox "Password (optional) More information about your password requirements Generate password"
```

and for `NumberField` with a contextual help:

```
textbox "Age (optional) More information"
```

Not specific to this component and not what #2498 reports. Fixing it means moving `Label`'s tunnelled actions out of the `<label>` element, which changes the root element of every form field's label — its own PR, with a cross-version run and a `ref`-contract decision. Happy to file it.

## Checklist

- [x] `pnpm lint` is clean and the affected browser, unit, compile and visual tests pass
- [x] **Generated code is committed** (`git diff` empty after `pnpm build`)
- [x] User-facing strings added to **both** `de-DE` and `en-US` locale files
- [x] Docs: no public API changed; no visual change, so no snapshot update


ins0:
Filed the `Label` defect from the PR description as #3235.

mfal (approved):

